PR #1, opened 2017-02-07
Made fvtr.sh able to print more detailed errors

denilsonm:
Functionality was added to fvtr.sh in order to make it possible to print logs of tests that failed easily. This was done through the addition of the subcommand "report" to fvtr.sh.

### DIFF
--- a/fvtr/fvtr.sh
+++ b/fvtr/fvtr.sh
@@ -35,6 +35,8 @@ CONFIGFILE=""
 
 function usage () {
 	echo "Usage: fvtr [-fl] <configfile> [test1 test2 ...]"
+	echo "       fvtr report"
+	echo ""
 	echo "       -f Force the test(s) to be run"
 	echo "       -l Output is sent to the console"
 
@@ -43,8 +45,14 @@ function usage () {
 	echo "                to run all tests by default."
 	echo "       Example: fvtr -f ~/advance-toolchain/config.at5.0.sles10 "
 	echo ""
-	echo "   or: fvtr report"
-	echo "       the report subcommand outputs the logs of failed tests"
+	echo "       report - prints content of log files that failed. This"
+	echo "                command does not run tests again. Instead, it"
+	echo "                only takes the logs of the last test that was"
+	echo "                ran with fvtr.sh in order to print relevant"
+	echo "                information. It's also important to note that"
+	echo "                this implies that a test must be necessarily"
+	echo "                ran without the -l option in order to enable"
+	echo "                later usage of the report command."
 	exit 0
 }
 
@@ -71,10 +79,21 @@ then
 		exit 1
 	fi
 
+	printcount=0
+
 	logname=$(awk 'NR==1' report.log)
 	while read line
 	do
 		testname=$(awk '{print $1}' <<< "$line")
+
+		# If no tests fail, a blank line is left in report.log,
+		# which we should ignore.
+		if [[ $testname = "" ]];
+		then
+			continue
+		else
+			printcount=$(($printcount+1))
+		fi
 
 		# Here we generate enough '*' so that "Log file content"
 		# lines have the same length, for readability purposes
@@ -85,6 +104,11 @@ then
 		cat "./$testname/$logname"
 		echo ""
 	done <<< "$(tail --lines=+2 report.log)"
+
+	if [[ $printcount = 0 ]];
+	then
+		echo "No tests failed during the last run of the testsuite."
+	fi
 
 	exit 0
 fi
--- a/fvtr/fvtr.sh
+++ b/fvtr/fvtr.sh
@@ -42,6 +42,9 @@ function usage () {
 	echo "       tests -  a list of tests to run (optional).  It is going"
 	echo "                to run all tests by default."
 	echo "       Example: fvtr -f ~/advance-toolchain/config.at5.0.sles10 "
+	echo ""
+	echo "   or: fvtr report"
+	echo "       the report subcommand outputs the logs of failed tests"
 	exit 0
 }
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -847,7 +847,7 @@ $(RCPTS)/fvtr.rcpt: $(RCPTS)/package.rcpt
 	   AT_WD=$(AT_WD) \
 	   AT_PREVIOUS_VERSION=$(AT_PREVIOUS_VERSION) \
 	   AT_DIR_NAME=$(AT_DIR_NAME) \
-	   ./fvtr.sh -f \
+	   ./fvtr.sh -f -d \
```
tuliom: I don't think this is a good idea.
It's going to print too much unnecessary data.

fvtr/fvtr.sh
```diff
@@ -37,6 +38,7 @@ function usage () {
 	echo "Usage: fvtr [-fl] <configfile> [test1 test2 ...]"
 	echo "       -f Force the test(s) to be run"
 	echo "       -l Output is sent to the console"
+	echo "       -d Print detailed log for errors"
```
tuliom: What is the difference between -d and -l?

tuliom:
Regardless of the comments I have, I think this commit needs a better commit message too.
In other words, try to improve your commit message by answering to the question: why should this commit be accepted?

Makefile
denilsonm: I made it this way because the requirement is that the builtbot should print detailed logs specifically for tests that failed. The alternative to placing -d as a default in the makefile would be creating a new option for make test that makes it pass the -d option to fvtr.sh, and then the buildbot passes the appropriate option to make test by default. Would that be preferable? Other suggestions?

fvtr/fvtr.sh
denilsonm: -l sends absolutely all logs to console. -d only sends logs of tests that actually failed. I did notice that I made a mistake, though: -l and -d should be mutually exclusive. I'll fix that in the next commit.

Makefile
tuliom: That's a good idea.
It's important that a buildbot requirement doesn't affect an user output.

fvtr/fvtr.sh
```diff
@@ -62,13 +65,111 @@ runall()
...
+
+	for param in "$@"
+	do
+		if [[ $param = "all" ]];
+		then
+			OUTPUTALL=ON
+		elif [[ $param = "errors" ]];
+		then
+			OUTPUTERR=ON
+		elif [ -e "$param" ] && [ -d "$param" ];
```
rzinsly: Test with -d is sufficient.

fvtr/fvtr.sh
```diff
@@ -62,13 +65,111 @@ runall()
...
+
 # We need at least the config file.
 if [[ ${#} -lt 1 ]]; then
 	usage
 	exit 1
 fi
 
-while getopts ":lfd" Option;
+
+
```
rzinsly: I don't believe more lines are necessary.

fvtr/fvtr.sh
```diff
@@ -62,13 +65,111 @@ runall()
...
+	then
+		# If the user requested only logs for tests that failed, we
+		# have to perform an extra check to see which tests failed
+		while read line
+		do
+			testname=$(awk '{print $1}' <<< "$line")
+			teststatus=$(awk '{print $2}' <<< "$line")
+
+			if [[ $teststatus = "failed"
+			   && $(echo "$OUTPUTLIST" | grep -c -w "$testname") = "0" ]];
```
rzinsly: Do we need this test?
Will report.log have duplicated entries?

fvtr/fvtr.sh
```diff
@@ -197,20 +306,27 @@ do
...
+						echo -e "$i\tpassed" >> report.log
+					fi ;;
 				"38")
 					echo -e "\tignored"
-					total_ignored=$(expr ${total_ignored} + 1) ;;
+					total_ignored=$(expr ${total_ignored} + 1)
+
+					if [[ $LOUD = "OFF" ]];
+					then
+						echo -e "$i\tignored" >> report.log
```
rzinsly: You don't need to duplicate code, you can always print the $i with "echo -n" and only print the status of failed tests.

fvtr/fvtr.sh
```diff
@@ -42,6 +42,9 @@ function usage () {
 	echo "       tests -  a list of tests to run (optional).  It is going"
 	echo "                to run all tests by default."
 	echo "       Example: fvtr -f ~/advance-toolchain/config.at5.0.sles10 "
+	echo ""
+	echo "   or: fvtr report"
+	echo "       the report subcommand outputs the logs of failed tests"
```
tuliom: When we created this function, we tried to follow the same output style of other programs.
This addition isn't following that style.  It would be interesting if it could follow it.
However, I don't think this change should block this patch.
When you make that change, please explain the difference between -l and report.

denilsonm: @tuliom Fixed.

fvtr/fvtr.sh
```diff
@@ -35,16 +35,23 @@ CONFIGFILE=""
 
 function usage () {
 	echo "Usage: fvtr [-fl] <configfile> [test1 test2 ...]"
+	echo "       fvtr report"
+	echo ""
 	echo "       -f Force the test(s) to be run"
 	echo "       -l Output is sent to the console"
-
+
```
rzinsly: Trailing whitespace here.

fvtr/fvtr.sh
```diff
@@ -35,16 +35,23 @@ CONFIGFILE=""
...
+	echo ""
 	echo "       -f Force the test(s) to be run"
 	echo "       -l Output is sent to the console"
-
+	
 	echo "       '<configfile>' path to the config file used to build AT"
 	echo "       tests -  a list of tests to run (optional).  It is going"
 	echo "                to run all tests by default."
 	echo "       Example: fvtr -f ~/advance-toolchain/config.at5.0.sles10 "
-	echo ""
```
rzinsly: This line you could keep to improve readability.

fvtr/fvtr.sh
```diff
@@ -35,16 +35,23 @@ CONFIGFILE=""
...
-	echo "   or: fvtr report"
-	echo "       the report subcommand outputs the logs of failed tests"
+	echo "       report - prints content of log files that failed. This"
+	echo "                command does not run tests again. Instead, it"
+	echo "                only takes the logs of the last test that was"
+	echo "                ran with fvtr.sh in order to print relevant"
+	echo "                information. It's also important to note that"
+	echo "                this implies that a test must be necessarily"
+	echo "                be ran with the -l option in order to enable later"
```
rzinsly: This line have more than 79 characters (see https://www.gnu.org/prep/standards/standards.html#Formatting).

rzinsly: You mean **without** the -l, right?

denilsonm: @rzinsly You're right, I'll fix it.

denilsonm: @rzinsly This line is actually only 74 characters long.

PR #9, opened 2017-02-17
Update python for AT 10.0

wainersm:
Updated python to version 3.5.3.

Changes in the patches:
- Removed python-3.5.1-fix_integer_overflow.patch: fix has been merged
  into 3.5.3 (see http://bugs.python.org/issue26171);
- New python-3.5.3-getlib64s1.patch rebased old lib64 patch to python
  3.5.3.

Signed-off-by: Wainer dos Santos Moschetta <wainersm@linux.vnet.ibm.com>

### DIFF
--- a/configs/10.0/packages/python/sources
+++ b/configs/10.0/packages/python/sources
@@ -20,9 +20,9 @@
 #
 
 ATSRC_PACKAGE_NAME="Python"
-ATSRC_PACKAGE_VER=3.5.1
+ATSRC_PACKAGE_VER=3.5.3
 ATSRC_PACKAGE_LICENSE="Python Software Foundation License 2"
-ATSRC_PACKAGE_DOCLINK="http://docs.python.org/release/3.4/"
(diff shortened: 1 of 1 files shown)


configs/10.0/packages/python/sources
```diff
@@ -41,24 +41,17 @@ ATSRC_PACKAGE_BUNDLE=toolchain_extra
 atsrc_get_patches ()
 {
 	at_get_patch \
-		https://raw.githubusercontent.com/powertechpreview/powertechpreview/26747dc0bfb0f8f30de9ebad78847a96d3191ba5/Python%20Fixes/python-3.5.1-getlib64s1.patch \
-		15ffa7f1c39763b7bf3cca0fd70bf421 || return ${?}
-
+		https://raw.githubusercontent.com/powertechpreview/powertechpreview/master/Python%20Fixes/python-3.5.3-getlib64s1.patch \
```
rzinsly: It' s better to get the patch from the commit ID instead of from master, this way you can guarantee its integrity.
If, for example, we remove the patch an old build will continue to work.

PR #12, opened 2017-02-28
Updated update_revision.sh to use GitHub

denilsonm:
Gerrit-related features were removed. After updating the revision on the local sources file, the script now:
- Checks connection to GitHub
- Finds out if the sources file with the latest revision has been previously pushed
- Commits and pushes to the user's fork

update_revision.sh now only takes as parameters the sources file path and the user's GitHub account. It is assumed that whoever is running the script is capable of pushing to the advance-toolchain fork of the provided user without any password.

As a side note, I also corrected some typos and weird indentation that were on the original file, but if any of them were intentional please let me know so I can undo what I did.

### DIFF
--- a/scripts/utilities/update_revision.sh
+++ b/scripts/utilities/update_revision.sh
@@ -16,32 +16,30 @@
 #
 # This script opens a config/source file, reads all available repositories
 # and change the ATSRC_PACKAGE_REV to the latest revision.
-# Afterwards it attempts to send the change for review, according to the
-# following rules:
-#  - If there isn't any pending change, then it submits a new commit.
-#  - If there's a pending change and its Code-Review or Verified label is set
-#  "-1", then it aborts.
-#  - If there's a pending change without "-1" label, then it submit a new
-#  commit but reuses the Change-Id. i.e. it updates the pending change with a
-#  new patch set.
+# Afterwards it checks if the sources file on the remote topic branch is
+# different from the sources file in the local topic branch. In case it's
+# not, there's nothing to be done. Otherwise, the script commits and
+# pushes to the remote topic branch on the advance-toolchain fork of the
+# provided user.
 #
-# Remarks: This operation creates another file, so links are discarded to avoid
-# affecting previous releases.
+# Remarks:
+# - This script uses SSH to push to GitHub, therefore you need
+# the appropriate keys of the provided GitHub user.
+# - Since this script automatically switches branches, it's
+# recommended unstaging or stashing changes before running it.
 #
 # Usage:
-#    update_revision <config/package/source> <gerrit_server> <gerrit_port>
+#    update_revision <config/package/source> <github_user>
 
-# This script requires 1 parameter.
-if [[ ${#} -ne  3 ]]; then
-	echo "update_revision.sh expects 3 parameters.";
+# This script requires 2 parameters.
+if [[ ${#} -ne  2 ]]; then
+	echo "update_revision.sh expects 2 parameters.";
 	return 1;
 fi
 
 # TODO: adjust following variables.
-# Assume user access gerrit password-lessly
-GERRIT_USER=$(whoami)
-GERRIT_SERVER=${2}
-GERRIT_PORT=${3}
+# Assume user access GitHub password-lessly
+GITHUB_USER=${2}
 
 source ${1};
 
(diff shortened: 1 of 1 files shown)


scripts/utilities/update_revision.sh
```diff
@@ -139,61 +108,58 @@ port ${GERRIT_PORT} with user \"${GERRIT_USER}\"."
...
+	echo -e ${msg} | git commit -F -
 
-	# Finally send to gerrit
+	# Finally send to GitHub
 	# Use topic branch to keep track of changes
-	print_msg 2 "Sending commit to Gerrit"
-        git push ssh://${GERRIT_USER}@${GERRIT_SERVER}:${GERRIT_PORT}\
-/advance-toolchain HEAD:refs/for/master%topic=${topic}
+	print_msg 2 "Sending commit to GitHub"
+	git push --force ${target_remote} HEAD:${target_branch}
```
tuliom: In Gerrit, when you push a patch to for/master, that means you're creating pull request.
On Github, the steps are different: after pushing the patch, it's necessary to open a pull request too.

scripts/utilities/update_revision.sh
```diff
@@ -139,61 +108,58 @@ port ${GERRIT_PORT} with user \"${GERRIT_USER}\"."
...
-		print_msg 0 "Reuse Change-Id of pending change on Gerrit"
-	else
-		print_msg 0 "Generate commit with a new Change-Id"
-	fi
-        echo -e ${msg} | git commit -F -
+	echo -e ${msg} | git commit -F -
 
-	# Finally send to gerrit
+	# Finally send to GitHub
 	# Use topic branch to keep track of changes
```
rzinsly: The new code doesn't use topics.

scripts/utilities/update_revision.sh
```diff
@@ -94,13 +154,57 @@ send_to_review ()
...
 
-	# Check connection to GitHub
+	# Check connection to GitHub or Gerrit
 	#
-	print_msg 2 "Checking connection to GitHub."
-	ssh git@github.com > /dev/null 2>&1
-	if [[ $? -ne 1 ]]; then
-		print_msg 0 "Cannot connect to GitHub."
-		return 1
+	if [[ ${service} = "github" ]];
```
rzinsly: It would be better to have a function for each service, this generic send_to_review function could have the common code and call the specific function.

scripts/utilities/update_revision.sh
```diff
@@ -112,12 +216,42 @@ send_to_review ()
...
+		gitdir=$(git rev-parse --git-dir)
+		if [[ ! -f ${gitdir}/hooks/commit-msg ]]; then
+			scp -p -P ${GERRIT_PORT} ${GERRIT_USER}@${GERRIT_SERVER}:\
+hooks/commit-msg ${gitdir}/hooks/
+		fi
+	fi
+
 	# Here we check if the new sources file we generated by updating the
 	# revision is different from the sources file at the last commit
 	# on the appropriate branch.
```
rzinsly: The following code will run even if the user choose gerrit, it shouldn't.

scripts/utilities/update_revision.sh
```diff
@@ -151,12 +285,53 @@ send_to_review ()
...
+	# Now we send the commit to the selected service
 	# Use topic branch to keep track of changes
-	print_msg 2 "Sending commit to GitHub"
-	git push --force ${target_remote} HEAD:${target_branch}
+	if [[ ${service} = "github" ]];
+	then
+		print_msg 2 "Sending commit to GitHub"
+		git push --force ${target_remote} HEAD:${target_branch}
+
+		pulljson=$(cat <<EOF
```
rzinsly: This is only needed in case of a $GITHUB_TOKEN exists.

scripts/utilities/update_revision.sh
```diff
@@ -151,12 +285,53 @@ send_to_review ()
...
+		)
+
+		if [[ ! -z "$GITHUB_TOKEN" ]];
+		then
+			pulltxt=$(curl -si \
+			https://api.github.com/repos/advancetoolchain/advance-toolchain/pulls \
+			-H "Authorization: token $GITHUB_TOKEN"\
+			--data "$pulljson")
+
+			if [[ $(echo "$pulltxt" | grep -c 'Status: 201 Created') -eq 0 ]];
```
rzinsly: It could print an error message in case of errors.

PR #13, opened 2017-03-01
Update GDB on AT 10.0 and next

tuliom:
Update GDB to version 7.12-2039d74e780d.


configs/10.0/packages/gdb/sources
```diff
@@ -19,8 +19,8 @@
 # =================================
 #
 ATSRC_PACKAGE_NAME="GDB (GNU Debugger)"
-ATSRC_PACKAGE_VER=7.11.1
-ATSRC_PACKAGE_REV=140bf80050b3
+ATSRC_PACKAGE_VER=7.12
```
RajalakshmiS: @tuliom Should it be 7.12.1 ?

tuliom: No, because this commit is not from the 7.12 branch.  It's actually from what will become 7.13.

PR #18, opened 2017-03-03
Improved remote-build.exp to support custom directories

denilsonm:
The script is now capable of copying and building AT on custom directories, through the usage of the optional --dir=<something> parameter.

### DIFF
--- a/remote-build.exp
+++ b/remote-build.exp
@@ -41,17 +41,6 @@ if {[file exists ${conf}]} {
 	source ${conf}
 }
 
-# Define default value of variables if not previously defined
-# Directory used by the script to save its files.
-set_def tray_dir ~/tmp/at-build-tray
-# Directory where AT will be installed (aka DESTDIR).
-set_def destdir "\${HOME}/opt"
-set_def default_timeout 30; # 30 seconds
-
-# Directory where AT is installed
-# Before start the building, all AT installs with the same version of that
-# specified in the parameters are removed
-set at_dest ~/opt
 set prompt "(%|>|\#|\\\$)\[\[:space:\]\]*$"
 
 # Request passwords once per server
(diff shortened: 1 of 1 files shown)


remote-build.exp
```diff
@@ -41,13 +41,6 @@ if {[file exists ${conf}]} {
...
-# Define default value of variables if not previously defined
-# Directory used by the script to save its files.
-set_def tray_dir ~/tmp/at-build-tray
-# Directory where AT will be installed (aka DESTDIR).
-set_def destdir "\${HOME}/opt"
-set_def default_timeout 30; # 30 seconds
-
 # Directory where AT is installed
 # Before start the building, all AT installs with the same version of that
 # specified in the parameters are removed
```
rzinsly: You could also remove the at_dest declaration as it seems it's not used anywhere.

remote-build.exp
```diff
@@ -123,6 +128,13 @@ if { ! [file exists configs/${at_version}] } {
 puts "AT version: $at_version"
 puts "Servers: $servers"
 
+# Define default value of variables if not previously defined
+# Directory used by the script to save its files.
+set_def tray_dir "~/$basedir/tmp/at-build-tray"
```
rzinsly: IMHO we don't need to use the tmp dir, we can use ~/$basedir/at-build-tray directly.
This is just an opinion and should not block the patch.

tuliom: If you remove tmp/ from the tray_dir, it will change the default behavior of the script.
People need to be notified about this before making that change.

remote-build.exp
```diff
@@ -101,6 +100,12 @@ while { ${argi} < ${argc} } {
 				error "Could not parse ${arg}"
 			}
 		}
+		"--dir=[^ ]*" {
```
tuliom: What about --basedir?  It makes it clearer which variable will be changed.

remote-build.exp
```diff
@@ -123,6 +128,13 @@ if { ! [file exists configs/${at_version}] } {
 puts "AT version: $at_version"
 puts "Servers: $servers"
 
+# Define default value of variables if not previously defined
+# Directory used by the script to save its files.
+set_def tray_dir "~/$basedir/tmp/at-build-tray"
+# Directory where AT will be installed (aka DESTDIR).
+set_def destdir "~/$basedir/opt"
```
tuliom: This is also changing the behavior of the script.
People need to be notified about this before this change.

remote-build.exp
denilsonm: It's a possibility too, I just didn't think there was a need for a 1:1 correspondence between internal variable names and parameter names. I'll change that in the next commit.

remote-build.exp
denilsonm: @tuliom Aha, not really! I mean, I think. Apparently for bash "x/./y" and "x//y" mean the same as "x/y", and "." is actually the default value for $basename if the the user doesn't specify --dir :)

Also, I'd like to mention that I did test the behavior I mentioned here, but I'm not sure if it's exactly... "canonical". In case there's a more standard way to do it, please let me know.

remote-build.exp
denilsonm: @rzinsly Okay, I'll remove it in the next commit.

remote-build.exp
tuliom: I didn't mean to say "default behavior", just "behavior".
Especially if "tmp/" is removed from tray_dir.

remote-build.exp
rzinsly: I understand and I'm ok with keeping the tmp dir if anyone bothers.

remote-build.exp
tuliom: I still think having a 1:1 correspondence would help understanding what each option is doing.

PR #20, opened 2017-03-07
Created AT 11.0 from next

denilsonm:
AT 11.0 was created from the next configset.


configs/11.0/packages/gdb/sources
```diff
@@ -0,0 +1,42 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# GDB source package and build info
+# =================================
+#
+ATSRC_PACKAGE_NAME="GDB (GNU Debugger)"
+ATSRC_PACKAGE_VER=7.11.1
```
tuliom: I think AT 10.0 and next already have a newer GDB.
I'd just symlink this file to AT 10.0.

configs/11.0/packages/glibc/sources
```diff
@@ -0,0 +1,67 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# GLIBC source package and build info
+# ===================================
+#
+
+ATSRC_PACKAGE_NAME="GNU C Library"
+ATSRC_PACKAGE_VER=2.26
```
tuliom: Same problem as GCC.
I'd like this to AT next while glibc 2.26 doesn't have its own branch.

configs/11.0/packages/gcc/sources
```diff
@@ -0,0 +1,57 @@
...
+
+# GCC source package and build info
+# =================================
+#
+
+ATSRC_PACKAGE_NAME="GCC (GNU Compiler Collection)"
+ATSRC_PACKAGE_SUBPACKAGE_NAME_0="GNU Standard C++ Library v3 (Libstdc++-v3)"
+ATSRC_PACKAGE_SUBPACKAGE_NAME_1="GNU Libgomp"
+ATSRC_PACKAGE_VER=7.0.1
+ATSRC_PACKAGE_REV=70d4d06ab769
```
rzinsly: That's a good idea!

configs/11.0/base.mk
```diff
@@ -0,0 +1,54 @@
...
+#
+# - AT_PREVIOUS_VERSION Previous compatible AT version supported (It may be
+#                       further restricted by distro specific settings)
+#
+AT_NAME := at
+AT_MAJOR_VERSION := 11.0
+AT_REVISION_NUMBER := 0
+AT_INTERNAL := alpha
+AT_PREVIOUS_VERSION := 10.0
+AT_DIR_NAME := at-next-$(AT_MAJOR_VERSION)-$(AT_REVISION_NUMBER)-$(AT_INTERNAL)
```
rzinsly: You need to remove this line.

configs/11.0/packages/gdb/sources
```diff
@@ -1,42 +0,0 @@
-#!/usr/bin/env bash
```
tuliom: Shouldn't this symlink be pointing to AT 10.0?

denilsonm: It should. I`ve corrected this through an amendment to the last commit.

tuliom:
Denilson, could you review if the author line of these patches are indeed formatted in the way you prefer?
I noticed it has only your first name.

rzinsly:
Bintutils and Python are outdated on those patches, they should reflect the files in the current master tree.

tuliom:
The changes LGTM, but I don't think we need 3 different commits for this.

PR #27, opened 2017-04-06
Added auto Signed-off-by to update_revision.sh

denilsonm:
Made the update_revision.sh script automatically add a Signed-off-by
line to commits to GitHub. Also corrected issues with private access
token scope detection.

### DIFF
--- a/scripts/utilities/update_revision.sh
+++ b/scripts/utilities/update_revision.sh
@@ -29,18 +29,23 @@
 # recommended unstaging or stashing changes before running it.
 #
 # Usage:
-#  update_revision <config/package/source> github <github_user> [PAT]
+#  update_revision <config/package/source> github <github_user> [PAT [signature]]
 #  update_revision <config/package/source> gerrit <gerrit_server> <gerrit_port>
 #
 #  The first parameter specifies which service to open a pull request
 #  on, after updating and commiting the sources file and committing.
 #
 #  If using GitHub, PAT stands for "Personal Access Token", and can be
-#  obtained on GitHub Setttings > Personal access tokens > Generate new
+#  obtained on GitHub Settings > Personal access tokens > Generate new
 #  token. For the current features of this script, only the public_repo
 #  scope is required to be enabled. In case you choose to omit your token,
 #  your topic remote branch will still be updated but no pull request
 #  will be opened.
+#    By default, this script will use your local git configuration to
+#  generate a signed-off-by line automatically for commits, which is
+#  required for contributing to AT. You may specify a custom signature
+#  as long as you also specify a PAT. Please follow the standard format,
+#  which is "your name here <your email here>".
 #
 #  If using Gerrit, after committing it attempts to send the change for
 #  review, according to the following rules:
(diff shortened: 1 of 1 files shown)


scripts/utilities/update_revision.sh
```diff
@@ -51,6 +51,12 @@
 #  commit but reuses the Change-Id. i.e. it updates the pending change with a
 #  new patch set.
 
+# IMPORTANT: if you change this script in any way, please update the
+# LAST_AUTHOR variable, using the format "your name <your email>".
+# This is used for the required Signed-off-by line in commits when
+# using GitHub.
+LAST_AUTHOR="Denilson Antonio Marques Junior <denimj@linux.vnet.ibm.com>"
```
rzinsly: This change doesn't make sense IMO, in this way the person who wants to use the scripts would need to change this file.
The patch needs to be signed off by the person who wrote it.

scripts/utilities/update_revision.sh
```diff
@@ -78,13 +84,30 @@ source ${1};
...
+		GIT_USER=$(git config --global user.name)
+		GIT_EMAIL=$(git config --global user.email)
+
+		if [[ "${GIT_USER}" = "" ]] || [[ "${GIT_EMAIL}" = "" ]];
+		then
+			echo "Error: Git's user.name or user.email not configured."
+			exit 1;
+		fi
+
+		GITHUB_SIGNATURE="${GIT_USER} <${GIT_EMAIL}>"
```
tuliom: Won't this overwrite what the user passed via command line?

denilsonm: It won't because it checks if the 5th argument is "" before assigning a new value to $GITHUB_SIGNATURE. But I'll replace the "${5}" = "" check on line 92 with "$GITHUB_SIGNATURE" = "" to make it more intuitive :)

scripts/utilities/update_revision.sh
```diff
@@ -78,13 +84,30 @@ source ${1};
 
 # If ATSRC_PACKAGE_CO is not defined, we skip the revision check.
 if ! [ -n "${ATSRC_PACKAGE_CO}" ] || ! [ -n "${ATSRC_PACKAGE_REV}" ]; then
-	return 0;
+	exit 0;
 fi
 
-# This function print a message with identation
+if [[ ${service} = "github" ]];
```
tuliom: Be careful with unquoted comparisons.
They can fail if `${service}` is undefined, a null string or a string whitespace string, e.g. `"  "`.
Even if the current code tests it earlier, belt-and-suspenders requires us to quote it.

scripts/utilities/update_revision.sh
```diff
@@ -324,7 +347,9 @@ public_repo scope enabled."
 
 	git add ${1}
 	local msg="Update ${pkg} on AT ${cfg}\n\
-Bump to revision ${2}\n\n"
+Bump to revision ${2}\n\n\
+\
+Signed-off-by: ${GITHUB_SIGNATURE}"
```
tuliom: You could make the same thing using `git commit -s`.

scripts/utilities/update_revision.sh
```diff
@@ -78,13 +84,30 @@ source ${1};
...
 # If ATSRC_PACKAGE_CO is not defined, we skip the revision check.
 if ! [ -n "${ATSRC_PACKAGE_CO}" ] || ! [ -n "${ATSRC_PACKAGE_REV}" ]; then
-	return 0;
+	exit 0;
 fi
 
-# This function print a message with identation
+if [[ ${service} = "github" ]];
+then
+	if [[ "${GITHUB_SIGNATURE}" = "" ]];
```
tuliom: I'm afraid this kind of comparisons only works on newer GNU Bash versions.  Usually, you can do the same thing using `-z "${GITHUB_SIGNATURE}"`.

denilsonm: Thanks for the suggestions! :D

PR #28, opened 2017-04-10
Use systemd to run watch_ldconfig on RPM based distros

rzinsly:
To be able to start watch_ldconfig after the runtime package is installed, this patch sets a systemd service for RPM based distros.

### DIFF
--- a/scripts/utilities/pkg_build_rpm.sh
+++ b/scripts/utilities/pkg_build_rpm.sh
@@ -89,6 +89,8 @@ function pkg_build_rpms()
 		rev_number=${at_revision_number}
 	fi
 
+	cp "${utilities}/cachemanager.service" "${rpmdir}"
+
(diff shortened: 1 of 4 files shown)


scripts/utilities/pkg_build_rpm.sh
```diff
@@ -134,7 +136,7 @@ function pkg_build_rpms()
 	# Final step to support debuginfo under /opt.
 	sed ${rpmdir}/find-debuginfo.sh -i -e "s@__AT_DEST__@${at_dest}@"
 
-	cp "${scripts}/utilities/split-debuginfo.sh" "${rpmdir}"
+	cp "${utilities}/split-debuginfo.sh" "${rpmdir}"
```
tuliom: Is this modification really expected to happen in this patch?

rzinsly: No, I removed it.

Issue #14, opened 2017-03-01
packages/zlib-1.2.11


grooverdan:
A recent spurt in releases has fixed a few things:

http://www.zlib.net/ChangeLog.txt

Still distributed via sf.

rzinsly:
Merged with PR #31.

Issue #24, opened 2017-04-05
Update Binutils on AT 10.0


rzinsly:
We need to update Binutils to the latest commit on binutils-2_27-branch to get the latest P9 patches on AT 10.0.

PR #31, opened 2017-04-11
Updates in zlib and binutils

youssefkanso:
Will close issue #14 and #24.


configs/10.0/packages/binutils/sources
```diff
@@ -21,7 +21,7 @@
 
 ATSRC_PACKAGE_NAME="GNU Binutils"
 ATSRC_PACKAGE_VER=2.27
-ATSRC_PACKAGE_REV=d72ca87af2af8
+ATSRC_PACKAGE_REV=b5d3ac25628b
```
tuliom: I see this revision has 11 characters instead of 12.
Wasn't we using 12 as a standard now?

rzinsly: Yes, we use 12.

rzinsly:
Merged, thanks!

Issue #19, opened 2017-03-07
binutils 2.28: internal error, aborting at ../../bfd/elf64-ppc.c:14952 in ppc64_elf_relocate_section


tuliom:
Fixed upstream on master with commit ID b1b07054c08eea65ba8d1a9f771796d686f337e5.
And on binutils-2_28-branch with commit ID 0d69827324f704b75ac7e5c2b39219cdd15e790b.

References:
- [Bug report](https://sourceware.org/bugzilla/show_bug.cgi?id=21224)

racardoso:
Closed via  dcea028

PR #36, opened 2017-04-20
Update binutils 2.28 for AT 11.0.

racardoso:
Updates binutils 2.28 to rev f85ae672d555. This fix a binutils
build error due to dynamic relocs (issue #19). Also includes
a patch to fix a incorrect pointer dereference at stabs.c.

Signed-off-by: Rogerio Alves Cardoso <rogealve@br.ibm.com>

### DIFF
--- a/configs/11.0/packages/binutils/sources
+++ b/configs/11.0/packages/binutils/sources
@@ -21,7 +21,7 @@
 
 ATSRC_PACKAGE_NAME="GNU Binutils"
 ATSRC_PACKAGE_VER=2.28
-ATSRC_PACKAGE_REV=7fa393306ed8
+ATSRC_PACKAGE_REV=f85ae672d555
 ATSRC_PACKAGE_LICENSE="GPL 2.0"
 ATSRC_PACKAGE_DOCLINK="http://sourceware.org/binutils/docs/"
 ATSRC_PACKAGE_RELFIXES=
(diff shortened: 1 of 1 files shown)


racardoso:
@rzinsly check fails for Ubuntu 16. It's 16.04?

rzinsly:
@racardoso yes, failed for ck_requires. Its a known failure and its not related to your patch.

configs/11.0/packages/binutils/sources
```diff
@@ -42,3 +42,20 @@ ATSRC_PACKAGE_PATCHES=
...
+		https://sourceware.org/ml/binutils/2017-02/msg00269.html \
+		fix-incorrect-indirection.patch 14 \
+		80e3c4062604f28f186289db0167b592 || return ${?}
+}
+
+atsrc_apply_patches ()
+{
+	#Fix error: comparison between pointer and zero character constant
+	#[-Werror=pointer-compare] in stabs.c
+	patch -p1 < fix-incorrect-indirection.patch
```
tuliom: If this patch fails to apply, this function will still return 0.

racardoso: @tuliom should `return ${?}` instead? There's other patches who do the same https://github.com/advancetoolchain/advance-toolchain/blob/master/configs/8.0/packages/binutils/sources#L58

tuliom: @racardoso, yes.  But I prefer the solution you adopted in the previous function using ` || return ${?}`.
That way you prevent to add a bug if someone else adds a new command between `patch` and `return`.

racardoso: @tuliom like ` patch -p1 < fix-incorrect-indirection.patch || return ${?}`? What about others places like 8.0? We should fix?

tuliom: Yes, like that.
I think we should fix the other places, but I don't think that's a requirement to get your patch accepted.

racardoso: @tuliom done.

PR #41, opened 2017-05-02
Update binutils on AT next

Cherbot:
Bump to revision 122a483d4f19


configs/next/packages/binutils/sources
```diff
@@ -21,7 +21,7 @@
 
 ATSRC_PACKAGE_NAME="GNU Binutils"
 ATSRC_PACKAGE_VER=2.28
```
tuliom: The version should be 2.28.51, which is read from bfd/version.m4.
However, I don't think this needs to be changed with this patch.

Issue #30, opened 2017-04-10
Userspace RCU doesn't pass the tests on SLES 12 x86_64 on AT 11.0


tuliom:
Needs to investigate and fix this error:

`
urcu: x86_64:  Start logging...
urcu: x86_64:  Testing 64 bit static linking...
urcu: x86_64:  Testing 64 bit dynamic linking...
urcu: x86_64:  Unrecognized error: /opt/at-next-11.0-0-alpha/bin/powerpc64le-linux-gnu-ld: warning: liburcu-common.so.4, needed by /opt/at-next-11.0-0-alpha/ppc64le/usr/lib/../lib64/liburcu.so, not found (try using -rpath or -rpath-link)
`

racardoso:
@tuliom can you give more details about how to reproduce this error? Like at version, make command.

tuliom:
@racardoso Sure! It's happening on AT next.  You can reproduce it if you use remote-build.exp on a SLES 12 x86_64 cross compiling to ppc64le.
Are you having trouble to reproduce it?

racardoso:
@tuliom I was trying using a stable version.

racardoso:
Still can't reproduce the problem:

Run `./remote-build.exp --buildarch=ppc64le next SLES_X86_64`

Fails test `check_package`:

```
Completed RPM packages for at12.0-0-alpha
Running FVTR for at12.0-0-alpha...
amino                   x86_64:         ignored
boost                   x86_64:         passed
ck_binaries             x86_64:         passed
ck_gold                 x86_64:         passed
ck_ldconfig             x86_64:         ignored
ck_ldds                 x86_64:         ignored
ck_libbits              x86_64:         passed
ck_lock_elision         x86_64:         ignored
ck_provides             x86_64:         ignored
ck_requires             x86_64:         passed
ck_tuned                x86_64:         ignored
dhrystone               x86_64:         ignored
gcc-builtins            x86_64:         passed
gccgo                   x86_64:         passed
gdb                     x86_64:         passed
gmp                     x86_64:         ignored
golang                  x86_64:         ignored
gotools                 x86_64:         ignored
libauxv                 x86_64:         passed
libdfp                  x86_64:         passed
libexpat                x86_64:         ignored
libhugetlbfs            x86_64:         passed
libvecpf                x86_64:         passed
mpfr                    x86_64:         ignored
openssl                 x86_64:         passed
oprofile                x86_64:         ignored
[...]

PR #46, opened 2017-05-05
Fix liburcu not found error on x86_64.

racardoso:
This commit fixes liburcu not found error detected by urcu test (issue #30)
when cross compile AT 11.0 on X86_64. The error was caused by a incorrect
library path searching on binutils when cross compiled on x86_64.

### DIFF
--- a/configs/11.0/packages/binutils/sources
+++ b/configs/11.0/packages/binutils/sources
@@ -49,6 +49,11 @@ atsrc_get_patches ()
 		https://sourceware.org/ml/binutils/2017-02/msg00269.html \
 		fix-incorrect-indirection.patch 14 \
 		80e3c4062604f28f186289db0167b592 || return ${?}
+
+	at_get_patch_and_trim \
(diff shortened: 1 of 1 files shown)


configs/11.0/packages/binutils/sources
```diff
@@ -49,11 +49,19 @@ atsrc_get_patches ()
 		https://sourceware.org/ml/binutils/2017-02/msg00269.html \
 		fix-incorrect-indirection.patch 14 \
 		80e3c4062604f28f186289db0167b592 || return ${?}
+
+	at_get_patch_and_trim \
+		https://sourceware.org/ml/binutils/2017-03/msg00173.html \
+		fix-urcu-rpath-error.patch 22 \
```
tuliom: Is it really RPATH?

racardoso: @tuliom Is not really a rpath. I give this name before based on this line of the reported error: `try using -rpath or -rpath-link`. The problem happens on dynamic linkage of libs on binutils due a wrong search of libs.

Issue #29, opened 2017-04-10
tcmalloc fails to compile ppc (32 bits) on AT 11.0 and next


tuliom:
The reported error is:

`In file included from src/stacktrace_impl_setup-inl.h:51:0,
                 from src/stacktrace.cc:129:
src/stacktrace_powerpc-linux-inl.h: In function ‘int GetStackTrace_ppc(void**, int, int)’:
src/stacktrace_powerpc-linux-inl.h:203:11: error: ‘siginfo_t’ does not name a type; did you mean ‘sigset_t’?
           siginfo_t info;
           ^~~~~~~~~
           sigset_t`

Needs to be investigated and fixed.

tuliom:
I updated the title because this is also happening when building ppc32 on a ppc64 system.

RajalakshmiS:
created PR https://github.com/gperftools/gperftools/pull/883

RajalakshmiS:
Merged

PR #52, opened 2017-05-15
Update tcmalloc on AT 11.0 (#29)

RajalakshmiS:
This patch updates tcmalloc to revision 50125d8 on AT 11.0.

Signed-off-by: Rajalakshmi Srinivasaraghavan  <raji@linux.vnet.ibm.com>

### DIFF
--- a/configs/11.0/packages/tcmalloc/sources
+++ b/configs/11.0/packages/tcmalloc/sources
@@ -1,1 +1,43 @@
-../../../10.0/packages/tcmalloc/sources
+#!/usr/bin/env bash
+#
+# Copyright 2017 IBM Corporation
+#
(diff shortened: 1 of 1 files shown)


RajalakshmiS:
@rzinsly This update is to make tcmalloc compile on ppc32. Not sure about libunwind failure.

rzinsly:
@RajalakshmiS the patch https://github.com/gperftools/gperftools/commit/7852eeb75b9375cf52a7da01be044da6e915dd08.patch is what fixes the libunwind failure, but by updating TCMalloc to the newer commit ID this patch doesn't need to be applied anymore.

In other words, you should remove the atsrc_get_patches and atsrc_apply_patches functions as we don't need to apply any patches.

Issue #45, opened 2017-05-05
sqlite3 python issue with AT9 and AT10


cristianomalossi:
I am experiencing an issue while loading sqlite3 from AT9 on a Minsky with RHEL7.3
 
```
> python3
Python 3.4.3 (default, Feb  1 2017, 14:32:30)
[GCC 5.3.1 20170120 (Advance-Toolchain-at9.0) [ibm/gcc-5-branch revision 244812 on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> os.path()
KeyboardInterrupt
>>> import sqlite
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: No module named 'sqlite'
>>> import sqlite3
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/at9.0/lib64/python3.4/sqlite3/__init__.py", line 23, in <module>
    from sqlite3.dbapi2 import *
  File "/opt/at9.0/lib64/python3.4/sqlite3/dbapi2.py", line 27, in <module>
    from _sqlite3 import *
ImportError: No module named '_sqlite3'
>>>
```
 
I have installed on RHEL the sqlite package:

> sqlite-devel
 
but I cannot do anything with pip to upgrade the python installation:

>  pysqlite is not supported on Python 3. When using Python 3, use the sqlite3 module from the standard library.

I think there is an issue with AT9 installation - probably that _sqlite3 file is missing.
The same issue appears with AT10.

You can also take a look at this page where they suggest to re-build python to fix it:
[No-module-named-sqlite3](http://stackoverflow.com/questions/1210664/no-module-named-sqlite3)

Can you please check?

Many thanks,

Cristiano MALOSSI

> Research Staff Member
> IBM Research - Zurich
> Foundations of Cognitive Solutions
> Säumerstrasse 4, 8803 Rüschlikon - Switzerland
> tel: +41 44 724 8616
> e-mail: acm@zurich.ibm.com
> www: http://ibm.biz/cristiano_malossi

rpsene:
by default the support is not activated. It is necessary to use --enable-loadable-sqlite-extensions during the configure process and install the libsqlite3-dev.

rpsene:
ubuntu@at-firstbuild:~/python_build/cpython$ ./python 
Python 3.4.6+ (default, May  8 2017, 14:08:33) 
[GCC 4.8.4] on linux
[...]

PR #59, opened 2017-05-23
Add support for sqlite on AT version of Python

rpsene:
This commit fixes the issue #45 which is
	related to the missing of sqlite3 in the
	python version built by AT. It adds the
	option --enable-loadable-sqlite-extensions
	during the confuration steps and includes
	the correct dependency of the devel package
	of sqlite for AT8.0 to AT Next.

### DIFF
--- a/scripts/utilities/update_revision.sh
+++ b/scripts/utilities/update_revision.sh
@@ -258,6 +258,11 @@ Bump to revision ${2}\n\n"
 #    $2 - revision ID
 send_to_github ()
 {
+	# Get AT config and package being updated.
+	# Expected a string like "<path-to-AT>/next/valgrind/source"
+	pkg=$(echo ${1} | awk -F "/" '{ print $(NF-1) }')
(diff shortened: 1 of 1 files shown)


fvtr/python/python.exp
```diff
@@ -105,6 +105,10 @@ set env(TMPDIR) "$FULLPATH/tmp[pid]"
 exec mkdir $env(TMPDIR)
 printit "TMPDIR=\"$env(TMPDIR)\""
 
+# Test for the issue #45
+printit "Executing sqlite3 importing test."
+exec $env(AT_DEST)/bin/python3 -c "import sqlite3"
```
tuliom: Isn't it possible to use test_sqlite?  It's already provided by the Python testsuite.
Besides that, are you sure this testcase will fail if the previous command fail?

rpsene: yes, the test case will brake.

PR #62, opened 2017-05-25
Add support for sqlite on AT version of Python

rpsene:
This commit fixes the issue #45 which is
	related to the missing of sqlite3 in the
	python version built by AT. It adds the
	option --enable-loadable-sqlite-extensions
	during the confuration steps and includes
	the correct dependency of the devel package
	of sqlite for AT8.0 to AT Next.

### DIFF
--- a/fvtr/python/python.exp
+++ b/fvtr/python/python.exp
@@ -91,7 +91,9 @@ if { "$pyver" == "2.6" } {
 	# Python unit tests are "embarrassingly parallel", at least up to 6 cores
 	# it seems
 	set CORES [expr min(6, [exec grep -c "processor" /proc/cpuinfo ])]
-
+        # Test for the issue #45
+        printit "Executing sqlite3 importing test."
+        exec $env(AT_DEST)/bin/python3 -c "import sqlite3"
 	# The following tests fail for python when run using FVTR so exclude
 	# them.
 	set spawn_cmd "spawn $env(AT_DEST)/bin/python3 $test_script \
--- a/configs/10.0/distros/centos-6.mk
+++ b/configs/10.0/distros/centos-6.mk
@@ -72,7 +72,7 @@ ifndef AT_DISTRO_REQ_PKGS
     AT_CROSS_PKGS_REQ :=
     AT_NATIVE_PKGS_REQ := \(ibm-java2-ppc64-sdk-5.0\|java-1.5.0-ibm-devel\) \
                           libxslt popt-devel qt-devel readline \
-                          readline-devel
+                          readline-devel sqlite-devel
     AT_COMMON_PKGS_REQ := zlib-devel ncurses-devel flex bison texinfo \
                           createrepo glibc-devel subversion cvs gawk autoconf \
                           rsync curl bc automake imake
--- a/configs/10.0/distros/centos-7.mk
+++ b/configs/10.0/distros/centos-7.mk
@@ -86,7 +86,7 @@ endef
 ifndef AT_DISTRO_REQ_PKGS
     AT_CROSS_PKGS_REQ :=
     AT_NATIVE_PKGS_REQ := libxslt docbook-style-xsl qt-devel \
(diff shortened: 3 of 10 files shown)


fvtr/python/python.exp
```diff
@@ -105,6 +105,10 @@ set env(TMPDIR) "$FULLPATH/tmp[pid]"
 exec mkdir $env(TMPDIR)
 printit "TMPDIR=\"$env(TMPDIR)\""
 
+# Test for the issue #45
+printit "Executing sqlite3 importing test."
+exec $env(AT_DEST)/bin/python3 -c "import sqlite3"
+
```
tuliom: In your previous pull request related to this patch, I made a suggestion and a question.

You did answer my question, correcting a wrong assumption from my part.
However, there were no mentions about the suggestion to reuse the test test_sqlite provided by the Python testsuite.
Don't you think this test would have a better coverage than just importing the sqlite3 module?

Notice that what I'm proposing here is to behave the same way as the other tests, e.g. test_gdb, test_site, test_httpservers, test_os, test_posix, test_compileall, etc.

tuliom: I've just found an error here: python3 is not available on all AT versions.  If we decide to use this test instead of reusing test_sqlite from the Python testsuite, this line has to be changed.

rpsene: What are you thoughts on increasing its coverage. Shouldn't the basic scenario regarding running a very small python call with its module be enough?

You mean adding this line inside the IF which corresponds to the python version? Like inside elseif { [string match 3.* $pyver] } ?

tuliom: > What are your thoughts on increasing its coverage. Shouldn't the basic scenario regarding running a very small python call with its module be enough?

You're just importing a module. There is no guarantee that the module is indeed working.
The best way we have to guarantee that is by running the test provided by the Python testsuite.

> You mean adding this line inside the IF which corresponds to the python version? Like inside elseif { [string match 3.* $pyver] } ?

That's one possible solution.  Just be careful to run it for each possible AT version that will support the sqlite module.

rpsene: According Python comments on sqlite test case (/Lib/test/test_sqlite.py):

import test.support

**# Skip test if _sqlite3 module not installed
test.support.import_module('_sqlite3')**

import unittest
import sqlite3
from sqlite3.test import (dbapi, types, userfunctions,
                                factory, transactions, hooks, regression,
                                dump)

def load_tests(*args):
    if test.support.verbose:
        print("test_sqlite: testing with version",
              "{!r}, sqlite_version {!r}".format(sqlite3.version,
                                                 sqlite3.sqlite_version))
    return unittest.TestSuite([dbapi.suite(), types.suite(),
                               userfunctions.suite(),
                               factory.suite(), transactions.suite(),
                               hooks.suite(), regression.suite(),
                               dump.suite()])

if __name__ == "__main__":
    unittest.main()


As we are now explicitly adding it, this test may run. So we don't need to worry about the test execution. Only move the test inside the python3 IF.

tuliom: > As we are now explicitly adding it, this test may run. So we don't need to worry about the test execution.

Looks like I made another mistake: test_sqlite is already being run with your patch.

>  Only move the test inside the python3 IF.

Yes!

fvtr/python/python.exp
```diff
@@ -91,7 +91,9 @@ if { "$pyver" == "2.6" } {
 	# Python unit tests are "embarrassingly parallel", at least up to 6 cores
 	# it seems
 	set CORES [expr min(6, [exec grep -c "processor" /proc/cpuinfo ])]
-
+        # Test for the issue #45
+          printit "Executing sqlite3 importing test."
+          exec $env(AT_DEST)/bin/python3 -c "import sqlite3"
```
tuliom: These 3 new lines aren't following the right indentation.

PR #65, opened 2017-05-28
Update gcc on AT next

Cherbot:
Bump to revision 8401372c12b7


rzinsly:
This patch breaks glibc build.

tuliom:
@rzinsly, the following patch should fix glibc and prevent this error:
https://sourceware.org/git/?p=glibc.git;a=commit;h=8fa11673d6b0feb3804eba6243e4979087e008a8

rzinsly:
@tuliom Thanks!

PR #85, opened 2017-06-07
Update gcc on AT next

Cherbot:
Bump to revision 5c672d0baa82


rzinsly:
This patch breaks GCC stage 2 build.

Issue #33, opened 2017-04-17
Some development files are still distributed in the runtime package


rzinsly:
Some files with extension .o, .la, .spec and .gox are still being distributed in the runtime package, they should be moved to the devel package.

rpsene:
Commit 5c6f0c3f94d6b5bfa674a40f42785a9af9b7e980 fixes this bug.

PR #91, opened 2017-06-13
Fix the file filtering for devel and runtime

rpsene:
This commit fixes the issue #33  by
	fixing the filtering of the files
	extensions (.o, .la, .spec and .gox)
	which were being added in the runtime
	package.


rzinsly:
These changes should affect only AT >= 11.0, or the user will have conflicts while updating old AT versions.

Issue #74, opened 2017-05-30
Update Valgrind on AT 11.0


rzinsly:
Update Valgrind to version 3.12.0 on AT 11.0.

racardoso:
Closed via  35ab9c8

PR #99, opened 2017-06-28
Update Valgrind on AT 11.0.

racardoso:
Update Valgrind to version 3.12.0 on AT 11.0 (task #74).

### DIFF
--- a/configs/11.0/packages/valgrind/sources
+++ b/configs/11.0/packages/valgrind/sources
@@ -20,18 +20,18 @@
 #
 
 ATSRC_PACKAGE_NAME="Valgrind"
-ATSRC_PACKAGE_VER=3.11.0
+ATSRC_PACKAGE_VER=3.13.0
 # Don't update the Valgrind revision without updating the VEX revision too.
 # They have to be in sync.
-ATSRC_PACKAGE_REV=15905
-ATSRC_PACKAGE_VEX_REV=3223
+ATSRC_PACKAGE_REV=16446
+ATSRC_PACKAGE_VEX_REV=3386
 ATSRC_PACKAGE_LICENSE="GPL 2.0"
 ATSRC_PACKAGE_DOCLINK="http://valgrind.org/docs/"
 ATSRC_PACKAGE_RELFIXES=
 ATSRC_PACKAGE_VER_REV="${ATSRC_PACKAGE_VER}-${ATSRC_PACKAGE_REV}"
 ATSRC_PACKAGE_STR_VER="${ATSRC_PACKAGE_NAME} ${ATSRC_PACKAGE_VER}"
 ATSRC_PACKAGE_PRE="test -d valgrind-${ATSRC_PACKAGE_VER_REV}"
-ATSRC_PACKAGE_CO=([0]="svn co --revision ${ATSRC_PACKAGE_REV} --ignore-externals svn://svn.valgrind.org/valgrind/trunk valgrind-${ATSRC_PACKAGE_VER_REV}")
+ATSRC_PACKAGE_CO=([0]="svn co --revision ${ATSRC_PACKAGE_REV} --ignore-externals svn://svn.valgrind.org/valgrind/branches/VALGRIND_3_13_BRANCH/ valgrind-${ATSRC_PACKAGE_VER_REV}")
 # TODO: Downloading Valgrind and VEX in different steps is sub-optimal for
 # caching mechanisms.
 # VEX is part of the Valgrind code, but it's kept in a different repository.
@@ -40,7 +40,7 @@ ATSRC_PACKAGE_CO=([0]="svn co --revision ${ATSRC_PACKAGE_REV} --ignore-externals
 # synchronize its own revision with the VEX repository and whenever we
 # checkout a specific revision from trunk, it will always get the latest
 # revision from VEX, causing failures on Valgrind.
-ATSRC_PACKAGE_POST="svn co svn://svn.valgrind.org/vex/trunk valgrind-${ATSRC_PACKAGE_VER_REV}/VEX -r ${ATSRC_PACKAGE_VEX_REV}"
+ATSRC_PACKAGE_POST="svn co svn://svn.valgrind.org/vex/branches/VEX_3_13_BRANCH valgrind-${ATSRC_PACKAGE_VER_REV}/VEX -r ${ATSRC_PACKAGE_VEX_REV}"
 ATSRC_PACKAGE_SRC="${AT_BASE}/sources/valgrind-${ATSRC_PACKAGE_VER_REV}"
 ATSRC_PACKAGE_WORK=${AT_WORK_PATH}/valgrind
 ATSRC_PACKAGE_MLS=""
@@ -57,13 +57,6 @@ atsrc_get_patches ()
 		https://github.com/powertechpreview/powertechpreview/raw/be030afbf8d78018ab77aeea58092e03d483e8ce/Valgrind%20iTrace%20Patches/3.11/vg-3110-itrace.v1.tgz \
 		337042b8ea29d476b69d439c154d62f0 || return ${?}
 
-	# Patch to fix the register R2 being clobbered in asm.
-	# TODO: Remove it when Valgrind BZ #371668 is fixed.
-	at_get_patch_and_trim \
-                https://bugsfiles.kde.org/attachment.cgi?id=101860 \
-                fix_r2_clobber.patch 44 \
-                e9c2e01092042c0162fbc9f2a5931bb5 || return ${?}
-
 	return 0
 }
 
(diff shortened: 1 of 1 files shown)


configs/11.0/packages/valgrind/sources
```diff
@@ -20,11 +20,11 @@
 #
 
 ATSRC_PACKAGE_NAME="Valgrind"
-ATSRC_PACKAGE_VER=3.11.0
+ATSRC_PACKAGE_VER=3.12.0
 # Don't update the Valgrind revision without updating the VEX revision too.
 # They have to be in sync.
-ATSRC_PACKAGE_REV=15905
-ATSRC_PACKAGE_VEX_REV=3223
+ATSRC_PACKAGE_REV=16456
```
gftg85: This revision is from after the 3.13 branch was created:

```
$ svn log --revision 16419
------------------------------------------------------------------------
r16419 | sewardj | 2017-05-24 13:10:30 -0300 (Wed, 24 May 2017) | 2 lines

Create branches/VALGRIND_3_13_BRANCH as a copy of trunk r16418.

------------------------------------------------------------------------
```

Besides that, this revision is also newer than the creation of the tag 3.13.0:

```
$ svn log --stop-on-copy tags/VALGRIND_3_13_0
------------------------------------------------------------------------
r16448 | sewardj | 2017-06-16 04:27:45 -0300 (Fri, 16 Jun 2017) | 2 lines

Swizzle external.

------------------------------------------------------------------------
r16447 | sewardj | 2017-06-16 04:26:24 -0300 (Fri, 16 Jun 2017) | 2 lines

Create tags/VALGRIND_3_13_0 as a copy of branches/VALGRIND_3_13_BRANCH, r16446.

------------------------------------------------------------------------
```

Did you really want to update that far, or were you planning to update only to 3.12?

configs/11.0/packages/valgrind/sources
```diff
@@ -20,11 +20,11 @@
 #
 
 ATSRC_PACKAGE_NAME="Valgrind"
-ATSRC_PACKAGE_VER=3.11.0
+ATSRC_PACKAGE_VER=3.12.0
```
gftg85: Taking into account my comment on line 26, should you change this to 3.13.0?

configs/11.0/packages/valgrind/sources
racardoso: I get the wrong revision. I will check this.

configs/11.0/packages/valgrind/sources
racardoso: As @rzinsly suggest I will change to 3.13.

racardoso:
@rzinsly Is that right use 3.13.0. As @gftg85 suggest we should not use 3.14-pre-release or trunk@<revision> (eg.: trunk@16442) as AT_VERSION? What do you think?

rzinsly:
We should not use the 3.14-pre-release, but it's ok to get the 3.13 revision from trunk.

gftg85:
> We should not use the 3.14-pre-release, but it's ok to get the 3.13 revision from trunk.

What do you mean?

Revision 16419 creates the branch VALGRIND_3_13 from trunk.  Afterwards, some backports from trunk are applied to this branch until it gets maked as 3.13.0 in revision 16446.

If we get any revision from trunk it will not be equivalent to 3.13.0.

rzinsly:
> Revision 16419 creates the branch VALGRIND_3_13 from trunk. Afterwards, some backports from
> trunk are applied to this branch until it gets marked as 3.13.0 in revision 16446.

I understand now, so we should get the revision from 3.13.0 (16446).

gftg85:
We are still checking out from trunk (which might be fine), but I want to raise the concern that we should probably start checking out from the branch.

We could check both valgrind and vex from the branch with the following changes:

```
-ATSRC_PACKAGE_CO=([0]="svn co --revision ${ATSRC_PACKAGE_REV} --ignore-externals svn://svn.valgrind.org/valgrind/trunk valgrind-${ATSRC_PACKAGE_VER_REV}")
+ATSRC_PACKAGE_CO=([0]="svn co --revision ${ATSRC_PACKAGE_REV} --ignore-externals svn://svn.valgrind.org/valgrind/branches/VALGRIND_3_13_BRANCH/ valgrind-${ATSRC_PACKAGE_VER_REV}")
```

```
-ATSRC_PACKAGE_POST="svn co svn://svn.valgrind.org/vex/trunk valgrind-${ATSRC_PACKAGE_VER_REV}/VEX -r ${ATSRC_PACKAGE_VEX_REV}"
+ATSRC_PACKAGE_POST="svn co svn://svn.valgrind.org/vex/branches/VEX_3_13_BRANCH valgrind-${ATSRC_PACKAGE_VER_REV}/VEX -r ${ATSRC_PACKAGE_VEX_REV}"
```

What do you think?

configs/11.0/packages/valgrind/sources
```diff
@@ -20,11 +20,11 @@
...
 
 ATSRC_PACKAGE_NAME="Valgrind"
-ATSRC_PACKAGE_VER=3.11.0
+ATSRC_PACKAGE_VER=3.13.0
 # Don't update the Valgrind revision without updating the VEX revision too.
 # They have to be in sync.
-ATSRC_PACKAGE_REV=15905
-ATSRC_PACKAGE_VEX_REV=3223
+ATSRC_PACKAGE_REV=16446
+ATSRC_PACKAGE_VEX_REV=3398
```
gftg85: This is not the correct combination.

Revision 3398 in vex's trunk is newer than 16446 in valgrind's trunk.
Besides that, revision 3392 was not backported to branches/VALGRIND_3_13_BRANCH/VEX.

gftg85: If we keep checking out from trunk, the correct revision from vex is 3385, since it is the last revision that was backported to the branch:

```
svn log --stop-on-copy svn://svn.valgrind.org/vex/branches/VEX_3_13_BRANCH
------------------------------------------------------------------------
r3386 | sewardj | 2017-05-31 11:47:53 -0300 (Wed, 31 May 2017) | 2 lines

Merge, from trunk, r3385 (Handle x86 CET prefixes (32 bit only)).
[...]
```

racardoso: @gftg85 I think it should be nice using the branch instead the trunk but isn't better send it as a different task?

racardoso: @gftg85 There's another thing. We use ATSRC_PACKAGE_VER=3.13.0 following the pattern MAJOR.MINOR.PATCH and the branches uses the pattern REPOSITORY_MAJOR_MINOR_BRANCH. We can replace dots for underscore but what about the PATCH I should trim it or we can use ATSRC_PACKAGE_VER=3.13 instead?

gftg85: > @gftg85 I think it should be nice using the branch instead the trunk but isn't better send it as a different task?

IMO, the right time to decide is now, but I'll let the other developers weigh in.

> @gftg85 There's another thing for the branch use suggestion. We use ATSRC_PACKAGE_VER=3.13.0 following the pattern MAJOR.MINOR.PATCH and the branches uses the pattern REPOSITORY_MAJOR_MINOR_BRANCH. We can replace dots for underscore but what about the PATCH I should trim it or we can use ATSRC_PACKAGE_VER=3.13 instead?

You could just hardcode "branches/VALGRIND_3_13_BRANCH/" (instead of trunk) in the URL.  The branch name is never going to change anyway and AT 11, once release, is not suppose to change the branches it fetches the packages from. (This also means you can keep ATSRC_PACKAGE_VER=3.13.0).

rzinsly: > @gftg85 I think it should be nice using the branch instead the trunk but isn't better send it as a
> different task?

IMO it's a simple change and could be done here, but we can create another task and let someone else work on it.

> @gftg85 There's another thing for the branch use suggestion. We use ATSRC_PACKAGE_VER=3.13.0 following the pattern MAJOR.MINOR.PATCH and the branches uses the pattern REPOSITORY_MAJOR_MINOR_BRANCH. We can replace dots for underscore but what about the PATCH I should trim it or we can use ATSRC_PACKAGE_VER=3.13 instead?

3.13.0 is the version of the release, the branch will have all 3.13 releases that's why it doesn't have the last number.
So the name of the branch doesn't interfere in anything.

configs/11.0/packages/valgrind/sources
```diff
@@ -20,18 +20,18 @@
...
 
 ATSRC_PACKAGE_NAME="Valgrind"
-ATSRC_PACKAGE_VER=3.11.0
+ATSRC_PACKAGE_VER=3.13.0
 # Don't update the Valgrind revision without updating the VEX revision too.
 # They have to be in sync.
-ATSRC_PACKAGE_REV=15905
-ATSRC_PACKAGE_VEX_REV=3223
+ATSRC_PACKAGE_REV=16446
+ATSRC_PACKAGE_VEX_REV=3385
```
gftg85: This misses one commit from the branch.

When I mentioned that if we kept checking out from trunk, we should checkout 3385, that was only valid if we actually kept checking out from trunk.

Now that we are checking out from the 3.13 branch, we have to get the revision that backports 3385 into the branch. The following command and output tells us that we should get revision 3386 in the branch:

```
$ svn log --stop-on-copy svn://svn.valgrind.org/vex/branches/VEX_3_13_BRANCH
------------------------------------------------------------------------
r3386 | sewardj | 2017-05-31 11:47:53 -0300 (Wed, 31 May 2017) | 2 lines

Merge, from trunk, r3385 (Handle x86 CET prefixes (32 bit only)).
```

So, you have to change ATSRC_PACKAGE_VEX_REV to 3386.

OK with this change.

racardoso:
@gftg85 done.

Issue #77, opened 2017-05-30
Update TBB on AT 11.0


rzinsly:
Update TBB to version 2017 Update 6 on AT 11.0.

PR #100, opened 2017-06-28
Update TBB on AT 11.0.

racardoso:
Update TBB to version 2017 Update 6 on AT 11.0 (Task #77). TBB 2017
updates was moved to github. There's no new source packages on the
old link anymore. The old packages 4.3 (2016) are still avaliable.

### DIFF
--- a/configs/11.0/packages/tbb/sources
+++ b/configs/11.0/packages/tbb/sources
@@ -1,1 +1,71 @@
-../../../10.0/packages/tbb/sources
+#!/usr/bin/env bash
+#
+# Copyright 2017 IBM Corporation
+#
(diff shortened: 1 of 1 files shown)


racardoso:
@tuliom  That's weird. Maybe is because it's a new file (the older one was a link that points to 10.0). About your requests.
1. Ok I will change.
2. There's lots of packages who checkout from the repo. Are your sure it's better to get it from tar.gz? That was a question I made for @rzinsly and he think was better get it from repo.
3. The last release was update 7. Should I update it to release 7 instead release 6?

rzinsly:
@racardoso Tulio is suggesting to get the tarball provided by github from the repo, I agree.
And we should get the update 7.

Issue #78, opened 2017-05-30
Update Userspace RCU on AT 11.0


rzinsly:
Update Userspace RCU to version 0.9.3 on AT 11.0.

racardoso:
Closed via 8201381aa0d21cfff79b934ef7c510de1b51e20a

PR #102, opened 2017-06-28
Update Userspace RCU on AT 11.0

racardoso:
Update Userspace RCU to version 0.9.3 on AT 11.0 (Task #78).

### DIFF
--- a/configs/11.0/packages/liburcu/sources
+++ b/configs/11.0/packages/liburcu/sources
@@ -1,1 +1,35 @@
-../../../10.0/packages/liburcu/sources
+#!/usr/bin/env bash
+#
+# Copyright 2017 IBM Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


racardoso:
@tuliom  I did not understand this comment:

> If not, this should affect AT 10.0 as well as it's a minor update

 Why this will affect AT 10.0 too? AT 11.0 source file had a sylink to AT.10.0 source but I deleted the symlink and created a new source file for AT 11.0. As if we should update this to 0.10.0 instead 0.9.3 I think that @rzinsly can answer that better.

rzinsly:
As 0.10.0 is already available we should get it on AT 11.0.

gftg85:
>> If not, this should affect AT 10.0 as well as it's a minor update.

> Why this will affect AT 10.0 too?

@tuliom meant that, if we were just going to update to 0.9.3, we should update both AT 10 and AT 11, because 0.9.2 to 0.9.3 is just a minor update.  Notice that minor updates are allowed in point releases of AT (e.g. AT 10.0-3, AT 10.0-4).

On the other hand, updating from 0.9.2 to 0.10.0 is not allowed in point releases, so you have to remove the symlink (as you did in this newer version).

LGTM.

gftg85:
Hmm, actually, you must update the commit message.  It still reads 0.9.3.

racardoso:
@gftg85 Yes. I forgot to update commit message.

Issue #75, opened 2017-05-30
Update Libpfm on AT 11.0


rzinsly:
Update Libpfm to version 4.8.0 on AT 11.0.

racardoso:
Closed via efb7a7b

PR #103, opened 2017-06-28
Update Libpfm on AT 11.0.

racardoso:
Update Libpfm to version 4.8.0 on AT 11.0 (Task #75).


configs/11.0/packages/libpfm/sources
```diff
@@ -20,8 +20,8 @@
 #
 
 ATSRC_PACKAGE_NAME="Libpfm"
-ATSRC_PACKAGE_VER=4.7.1
-ATSRC_PACKAGE_REV=ebb8be7
+ATSRC_PACKAGE_VER=4.8.0
+ATSRC_PACKAGE_REV=f3c2293
```
tuliom: Should the revision follow the same pattern of the other packages, with 13 characters?

racardoso: I guess it should be 13 characters. I did not change it because it was using 7 so I left that way. But I will change it to 13.

rzinsly: Actually it is 12 characters.

gftg85: lgtm with this change.

racardoso:
@rzinsly Sorry I guess I did not see that Squash and Merge was disabled for this repo. I will rebase and squash those commits.

PR #105, opened 2017-07-03
Update glibc on AT next

Cherbot:
Bump to revision 8dc6133eff1c


tuliom:
This is causing a tcmalloc build failure.

tuliom:
The error message is:

```
In file included from src/stacktrace_impl_setup-inl.h:51:0,
                 from src/stacktrace.cc:129:
src/stacktrace_powerpc-linux-inl.h:56:9: error: 'ucontext' does not name a typ
e; did you mean 'ucontext_t'?
 typedef ucontext ucontext_t;
         ^~~~~~~~
         ucontext_t
```

Which is likely associated with [glibc bug #21457](https://sourceware.org/bugzilla/show_bug.cgi?id=21457).
This will require a patch to gperftools.

tuliom:
This patch was superseded by commit 4129b1d8ae06.

PR #107, opened 2017-07-05
Update gcc on AT next

Cherbot:
Bump to revision 967ec7427cf0


rzinsly:
The GCC update is being reviewed at #119.

PR #110, opened 2017-07-12
Use systemd in our cache manager system for DEB based distros

rzinsly:
On RPM based distros we use a  a systemd service instead of a cronjob to start our mechanism to automatic update AT's loader cache, these patches does the same for DEB based distros on AT >= 11.0.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1118,15 +1118,15 @@ $(RCPTS)/monitor.rcpt: $(RCPTS)/toolchain.rcpt
 	    group=$$( ls $(DYNAMIC_SPEC)/ | grep "toolchain\$$" ); \
 	    echo "$(AT_DEST)/bin/watch_ldconfig" \
 	          >> $(DYNAMIC_SPEC)/$${group}/ldconfig.filelist; \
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -1118,15 +1118,15 @@ $(RCPTS)/monitor.rcpt: $(RCPTS)/toolchain.rcpt
 	    group=$$( ls $(DYNAMIC_SPEC)/ | grep "toolchain\$$" ); \
 	    echo "$(AT_DEST)/bin/watch_ldconfig" \
 	          >> $(DYNAMIC_SPEC)/$${group}/ldconfig.filelist; \
-	    if [[ "$(pack-sys)" == "deb" || $${at_major_version%.*} -lt  9 ]]; then \
+	    if [[ $${at_major_version%.*} -lt  9 ]] || [[ "$(pack-sys)" == "deb" && $${at_major_version%.*} -lt 11 ]]; then \
```
tuliom: IMHO, the definition of which system to use should be in the distro.mk file.
Anyway, your patch is not adding a new behavior...

rzinsly: Agreed, I'll do that in another commit, thanks!

Issue #63, opened 2017-05-25
add an environment module


ThinkOpenly:
Given that the Advance Toolchain is installed in /opt, and using it effectively generally requires prepending to PATH (and MANPATH, INFOPATH), having an environment module to set those correctly, and remove those modifications trivially, would be a nice addition.  Something like:
```
$ echo $PATH;echo $MANPATH;echo $INFOPATH; which gcc
/usr/local/bin:/usr/bin:/bin:/usr/local/games:/usr/games
/usr/share/man

/usr/bin/gcc
$ module avail

---------------- /home/pc/modules ----------------
at10.0 at9.0
$ module load at10.0
$ echo $PATH;echo $MANPATH;echo $INFOPATH; which gcc
/opt/at10.0/bin:/usr/local/bin:/usr/bin:/bin:/usr/local/games:/usr/games
/opt/at10.0/share/man:/usr/share/man
/opt/at10.0/share/info
/opt/at10.0/bin/gcc
$ module unload at10.0
$ echo $PATH;echo $MANPATH;echo $INFOPATH; which gcc
/usr/local/bin:/usr/bin:/bin:/usr/local/games:/usr/games
/usr/share/man

/usr/bin/gcc
$

```

ThinkOpenly:
The environment-modules package is not always installed.  Do we want to add a dependency on that package, presumably from the -devel package, or create a new package for these tiny files?

ThinkOpenly:
I'm attaching an environment module for at9.0.  I had to rename it from "at9.0" to get it attached, but it should be shipped as "at9.0", so a user can load it with that shorter name.
[at9.0.txt](https://github.com/advancetoolchain/advance-toolchain/files/1045594/at9.0.txt)
It would be easy to just add that file somewhere in the AT -devel package, but to be most easily used, it would help for the package to require "environment-modules" (I think this works on all the supported distros), and place the file in an appropriate directory for system modules.  Unfortunately, the distros don't agree on any common directory.  Fortunately, the list of directories can be found in the MODULEPATH environment variable for all distros.  Unfortunately, MODULEPATH may not be defined if the package wasn't already installed.  :-/

ThinkOpenly:
Regardless of distro, there is a ".modulespath" file included in the package that defines the default MODULEPATH, so actual final placement could be customized based on the distribution (a bit ugly, but functional):

RHEL7.2:(environment-modules):
/usr/share/Modules/modulefiles
/etc/modulefiles

[...]

PR #112, opened 2017-07-13
Add support for environment modules

ThinkOpenly:
The Advance Toolchain is a good use-case for environment modules because
to effectively use it requires accurately setting a few independent
environment variables (PATH, MANPATH, INFOPATH).  Using an environment
module can do (and undo) this with single, simple commands:
$ module load at11.0
$ module unload at11.0

This patch, for AT 11 and higher:
- adds a dependency of "environment-modules" for the "-devel" package
- generates the environment module definition file automatically, based on
  the version of AT (including unique names for alpha and beta releases)
- includes the modulefile in the -devel package
- creates a symlink from the canonical system modulefiles directory to the
  AT modulefile, making the module visable with "module avail" and usable
  with "module load/unload/show/whatis"
- cleans up the symlink upon removal of the "-devel" package

AT_VER_REV_INTERNAL was used as the name of the environment module, as it
includes the version, release, and internal qualifier (like "alpha"),
which matches the normal installed path of the Advance Toolchain.
In the DEB-based processing, this variable had been modified to remove
any "." characters, presumably to appease "cron" or a similar external
dependency.  I reverted this modification through all versions, and
adjusted the individual usage to match the historical results where needed.


ThinkOpenly:
Fixes #63

configs/10.0/deb/monolithic/rules
```diff
@@ -62,7 +62,7 @@ binary-cp: build
 	# executed.
 	mkdir -p debian/tmp/etc/cron.d/
 	echo "@reboot root __AT_DEST__/bin/watch_ldconfig &" \
-	      > debian/tmp/etc/cron.d/__AT_VER_REV_INTERNAL___ldconfig
+	      > debian/tmp/etc/cron.d/__AT_VER_REV_INTERNAL_NODOT___ldconfig
```
tuliom: Where is ```__AT_VER_REV_INTERNAL_NODOT__``` defined?

ThinkOpenly: this patched added that in pkg_build_deb.sh

ThinkOpenly:
@tuliom said "Do we really want to start depending on the package environment-modules?"

That's certainly subjective.  What is the concern?  More difficult to resolve dependencies on non-connected systems?

Currently, "-devel" package (on Debian) already depends on guile-2.0-libs, libgc1c2 (>= 1:7.2d), liblzma5 (>= 5.1.1alpha+20110809), libncurses5 (>= 5.5-5~), libtinfo5.

It could be moved to a new package, although it would be extraordinarily tiny.  This might be a good time to discuss if an environment module would ever be shipped with advance-toolchain cross, and if so, how might that be packaged?  This basic one could be used without issue, and should be.  I'll open an issue for that.  But what about one that establishes aliases for the cross command set, so they can be invoked without changes to the makefile, for example?

configs/11.0/build.mk
```diff
@@ -1 +0,0 @@
-../10.0/build.mk
```
rzinsly: Seems we this file is symlinking to 10.0 on AT next too, we need to fix this by linking with AT 11.0.

rzinsly:
> Do we really want to start depending on the package environment-modules?
Does it make sense to pack the environment module in another package?

I prefer to not depend on it.

>It could be moved to a new package, although it would be extraordinarily tiny. This might be a good time to discuss if an environment module would ever be shipped with advance-toolchain cross, and if so, how might that be packaged? This basic one could be used without issue, and should be. I'll open an issue for that. But what about one that establishes aliases for the cross command set, so they can be invoked without changes to the makefile, for example?

I don't see a problem in provide a cross-compiler module but I wonder if would be used, I believe when people are working with cross it's more common to set variables like CC than use PATH.

ThinkOpenly:
Regarding the new dependency on environment-modules, there are a few options:
- keep it (this seems to be already decided against)
- change it to Recommends, or Suggests, neither of which would fail the installation if unavailable. Actual action depends on the resolver.
- move to a new package, like "advance-toolchain-devel-extras" (or something better) which maintains the dependency

ThinkOpenly:
@rzinsly said, "I don't see a problem in provide a cross-compiler module but I wonder if would be used, I believe when people are working with cross it's more common to set variables like CC than use PATH."
(I'm not sure why there isn't a way to reply to comments more directly...?  Anyway...)
As it stands, one has to either:

1. Make sure to override every tool invoked during the build with the equivalent tool from the Advance Toolchain cross compiler.  This _only_ works if the Makefile is suitably capable to support overrides.  If the lowest-level build steps are "gcc" and "ld" and not "$(CC)" and "$(LD)", then the Makefile must be edited.  And there can be a LOT to override for complex build environments.  The Advance Toolchain cross compiler provides "as", "c++", "cpp", "g++", "gcc", "gccgo", "gfortran", "ld", "strip", etc.

2. Set up symlinks for all of the tools invoked during the build, pointing to the equivalent tool provided by the Advance Toolchain (since the names are all different...I've see people do this).

It just seems easier to do "module load at11.0-cross", and have all of the aliases set with a single short command.

configs/11.0/build.mk
ThinkOpenly: There are two changes involved:
- pkg_build_deb.sh.  Before the change, this set __AT_VER_REV_INTERNAL__ from "at_ver_rev_internal", but without any dots.  This script is not release-dependent.  After the change, it sets __AT_VER_REV_INTERNAL__ directly from "at_ver_rev_internal", including the dots; and it sets __AT_VER_REV_INTERNAL_NODOT__ from "at_ver_rev_internal" without the dots.  Since the script is not release-dependent, all release-dependent files which used __AT_VER_REV_INTERNAL__ needed to change.  Thus the following additional changes.

- ./configs/7.1/deb/monolithic/rules
- ./configs/9.0/deb/monolithic/rules
- ./configs/10.0/deb/monolithic/rules
  all of these simply had changes to use the new variable:
--
-             > debian/tmp/etc/cron.d/__AT_VER_REV_INTERNAL___ldconfig
+             > debian/tmp/etc/cron.d/__AT_VER_REV_INTERNAL_NODOT___ldconfig
--

So, this change should work in all releases.

rzinsly: It's not your patch fault but it will break:
$ readlink configs/next/build.mk 
../10.0/build.mk

It should be ../11.0/build.mk

rzinsly:
> Regarding the new dependency on environment-modules, there are a few options:

    keep it (this seems to be already decided against)
    change it to Recommends, or Suggests, neither of which would fail the installation if unavailable. Actual action depends on the resolver.
    move to a new package, like "advance-toolchain-devel-extras" (or something better) which maintains the dependency

IMHO use Suggest seems a good option.

rzinsly:
> It just seems easier to do "module load at11.0-cross", and have all of the aliases set with a single short command.

Makes sense, seems even more helpful on cross than on native.

configs/11.0/build.mk
rzinsly: I got confused by where the replies are placed, are you talking about my comment regarding scripts/utilities/pkg_build_deb.sh?

If so the problem is that 'echo "${at_dest}/share/modules/modulefiles/${at_ver_rev_internal}"' puts the module file on old AT releases as well, dh_install will not find it (as it doesn't exist on AT < 11.0) thus breaking the build.

ThinkOpenly: This also breaks the -cross build.  I'll work on all of these issues and repush when ready.

tuliom:
> Regarding the new dependency on environment-modules, there are a few options:
> 
>     change it to Recommends, or Suggests, neither of which would fail the installation if unavailable. > Actual action depends on the resolver.

Changing it to Suggests seems a good alternative, but it won't work on RPM-based distros, unless they support RPM >= 4.12.

>     move to a new package, like "advance-toolchain-devel-extras" (or something better) which maintains the dependency

In case this is necessary, I suggest something more specific, e.g. "advance-toolchain-atX.X-environment-module".

tuliom:
In build.mk you say BUILD_ENVIRONMENT_MODULES should be 'yes' or 'no'.
But the variable was not created in older versions.
I didn't find any code marking it as 'no' if the definition isn't available.

Is it really safe to keep it this way?

tuliom:
There are trailing whitespaces in:
- configs/11.0/deb/monolithic/devel.prerm:20
- configs/11.0/specs/monolithic.spec:343 and 344

rzinsly:
This is currently being reviewed by https://github.com/advancetoolchain/advance-toolchain/pull/133

Issue #114, opened 2017-07-18
Re-enable libsanitizer on AT >= 11.0


tuliom:
libsanitizer had to be disabled on GCC stage 4 due to [GCC bug #81066](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=81066).

The fix has been backported to FSF's GCC 7 and libsanitizer should be enabled again on AT >= 11.0.

rpsene:
Fixed by PR #117

tuliom:
This patch broke gcc_1 on AT next.

tuliom:
The build failure on gcc_1 was not caused by this patch.  I'm closing this again.

PR #117, opened 2017-07-19
Re-enable libsanitizer on AT >= 11.0

rpsene:
Fix for #114    

    This commit re-enables libsanitizer on both versions
    of AT (native and cross) for versions greater then
    11.0. It affects the stages 2, 3, 4 and optimized of
    GCC build.

### DIFF
--- a/configs/11.0/packages/gcc/stage_4
+++ b/configs/11.0/packages/gcc/stage_4
@@ -36,7 +36,6 @@ atcfg_pre_configure() {
 atcfg_configure() {
 	if [[ "${cross_build}" == "no" ]]; then
 		# Configure command for native builds
-		# TODO: Re-enable libsanitizer when GCC bug #81066 is fixed.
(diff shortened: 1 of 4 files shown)


configs/11.0/packages/gcc/stage_4
```diff
@@ -36,8 +36,7 @@ atcfg_pre_configure() {
 atcfg_configure() {
 	if [[ "${cross_build}" == "no" ]]; then
 		# Configure command for native builds
-		#TODO: Enable gotools.
-		# TODO: Re-enable libsanitizer when GCC bug #81066 is fixed.
+		# TODO: Enable gotools.
```
tuliom: I know this is unrelated to your patch, but: isn't gotools already enabled?

rzinsly: Yes and the current tree already has this comment removed.

Issue #113, opened 2017-07-14
Update Expat on AT >= 11.0


rzinsly:
We need to update Expat to version 2.2.3 on AT 11.0 and next to fix  CVE-2017-9233.

racardoso:
Closed via  2e3fd9c

racardoso:
Renamed issue to AT >= 11.0 since it's the priority now. AT 10.0 and 9.0 should be on another task. They release 2.2.3 it's already include the entropy fix.

PR #123, opened 2017-07-26
Update Expat on AT >= 11.0.

racardoso:
Update Expat to version 2.2.2 on AT 10.0 and 11.0 to fix CVE-2017-9233 (task #113).

### DIFF
--- a/configs/11.0/distros/centos-6.mk
+++ b/configs/11.0/distros/centos-6.mk
@@ -1,1 +1,91 @@
-../../10.0/distros/centos-6.mk
+# Copyright 2017 IBM Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Distro dependent basic definitions for RedHat Enterprise Server 6
+# =================================================================
+# This Makefile include contains the definitions for the distro being used
+# for the build process. It contains kernel version infos and distro specific
+# sanity checks.
+
+# Informs the distro supported power archs
+AT_SUPPORTED_ARCHS := ppc64
+
+# You may force the BUILD_IGNORE_AT_COMPAT general condition by distro
+#BUILD_IGNORE_AT_COMPAT := yes
+
+# Kernel version to build toolchain against
+# - As this distro only supports one arch variation (ppc64), there is no
+#   need to conditionally define these versions based on arch.
+AT_KERNEL := 2.6.32     # Current distro kernel version for runtime.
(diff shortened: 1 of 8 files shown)


racardoso:
I will apply the patch to fix CVE-2017-9233 on 9.0 instead update to 2.2.2. It will be send on other PR.

racardoso:
@rzinsly The problem is that expat uses a getrandom() which is only available in glibc 2.25 or later which depends on  syscall(SYS_getrandom) which is available on Linux 3.17. We can workarround it by append `-DXML_POOR_ENTROPY` to `CPPFLAGS`

racardoso:
Dammit autools, I guess the problem is here:

https://github.com/libexpat/libexpat/blob/master/expat/configure.ac#L165

I have to move this `AC_CHECKING` outside the first `AC_CHECKING` because it makes no sense since on kernel > 3.17 we have `syscall(SYS_getrandom)`. In fact I think it's a mistake to put it inside the the other check. They're two separated check:

https://github.com/libexpat/libexpat/blob/master/expat/lib/xmlparse.c#L57

tuliom:
> I have to move this AC_CHECKING outside the first AC_CHECKING because it makes no sense since on kernel > 3.17 we have syscall(SYS_getrandom). In fact I think it's a mistake to put it inside the the other check.

AFAIU, this is checking if glibc supports getrandom, if it supports, use it the C wrapper.
If not, see if Linux supports the syscall without using the C wrapper.
If you move this out, it will always call getrandom directly without using the C wrapper.

racardoso:
@tuliom 

> If not, see if Linux supports the syscall without using the C wrapper.

Well but that's not working since it return false, even when Linux has support.

tuliom:
`ATSRC_PACKAGE_VER="R_2_2_3"`

This variable is used when generating the release notes.  We try our best to keep this value as human readable as possible.
I suggest to use the value 2.2.3 here

racardoso:
@tuliom About change the document link to github I agree I will change it. Now about `ATSRC_PACKAGE_VER="R_2_2_3"` It's because this string is used to get the package. I will check if I can change it.

tuliom:
@racardoso you can use the commit ID to download the source code.

_Sent from my Google Nexus 4 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

racardoso:
@tuliom That's the same case we agreed at TBB lib. That I was download the source code (using the commit) and you suggest that we should download the tar.gz instead.

tuliom:
@racardoso, no, I'm not suggesting to change the download method. Just replace the variable used.
You can still download the tarball directly from the commit ID, e.g. https://github.com/libexpat/libexpat/archive/97c6bd019900.tar.gz

racardoso:
@tuliom I was not aware about that option. I'll try it here.

racardoso:
@rzinsly I removed that by mistake when solve the merge conflicts. But the commit that update expat on AT 10.0 should not be there. I'll check it here.

racardoso:
@rzinsly Done I removed that commit who updates expat for AT 10.0. Should not be there. It will be done on another PR. For now I am only updated AT 11.0. 

The problem was when I first sent the PR (for AT 10.0) you was still working on the systemd problem. I forgot to rebase the commit. And when I rebase my branch for the Update AT >= 11.0 commit I forgot to delete this commit. Then I rebase/push --force and the things got messed.

PR #139, opened 2017-08-07
Update glibc on AT next

Cherbot:
Bump to revision 219dd320d69d


tuliom:
This patch looks good to me, but I think it should be merged after branching out AT 11.0 because this commit ID is already glibc 2.27.

rzinsly:
@tuliom done by https://github.com/advancetoolchain/advance-toolchain/commit/bc97e6f6339c6ab5f066d7dd5eb0ef452f4eda5e.

PR #133, opened 2017-08-01
Rely on the distro files to use the systemd service 

rzinsly:
This patch changes how we choose the mechanism to monitor AT's
loader cache by relying in a variable provided by the distro files.

This fixes the Ubuntu 14.04 build on AT >= 11.0.

PR #142, opened 2017-08-08
Systemd fixes

rzinsly:
This patch changes the names of the systemd files to avoid name clash
and work in all environments.

This PR depends on PR #133.

### DIFF
--- a/configs/9.0/specs/monolithic.spec
+++ b/configs/9.0/specs/monolithic.spec
@@ -191,10 +191,10 @@ systemd_preset=$(pkg-config --variable=systemdsystempresetdir systemd)
 mkdir -p ${RPM_BUILD_ROOT}/${systemd_unit}/
 sed -e s:__AT_VER_REV_INTERNAL__:%{at_ver_rev_internal}: \
     -e s:__AT_DEST__:%{_prefix}: %{_rpmdir}/cachemanager.service \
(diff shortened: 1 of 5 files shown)


configs/9.0/specs/monolithic.spec
```diff
@@ -191,10 +191,10 @@ systemd_preset=$(pkg-config --variable=systemdsystempresetdir systemd)
 mkdir -p ${RPM_BUILD_ROOT}/${systemd_unit}/
 sed -e s:__AT_VER_REV_INTERNAL__:%{at_ver_rev_internal}: \
     -e s:__AT_DEST__:%{_prefix}: %{_rpmdir}/cachemanager.service \
-    > ${RPM_BUILD_ROOT}/${systemd_unit}/%{at_ver_rev_internal}-cachemanager.service
+    > ${RPM_BUILD_ROOT}/${systemd_unit}/%{at_ver_alternative}-cachemanager.service
 mkdir -p ${RPM_BUILD_ROOT}/${systemd_preset}/
 echo "enable at*cachemanager.service" \
```
gftg85: Although not a blocker, would you also like to change this line so that only the same version gets enabled?

gftg85:
My previous comment is not a blocker. It could be addressed by another work item.

This patch looks good to me.

configs/9.0/specs/monolithic.spec
rzinsly: Makes sense.

PR #153, opened 2017-08-18
Fixes to the cache manager mechanism

### LINKED ISSUE #145: Fix the permission of the crontab file
The crontab file should have mode bits 644.
Having 664 may spam the syslog of some Linux distributions. Ubuntu 17.04 is known to be affected by this.

rzinsly:
This patch changes to 644 the mode of the cron file because in some
systems the script doesn't run with 664 permissions.
It also removes the redundancy of the systemd preset file name and
specify the version that the file enables.
Lastly, now the ubuntu packages don't fail in the post install if
the systemd service doesn't start by an unexpected behaviour like
what happens on docker.
This fixes #145 and #132.

Signed-off-by: Raphael Moreira Zinsly <rzinsly@linux.vnet.ibm.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1125,7 +1125,7 @@ $(RCPTS)/monitor.rcpt: $(RCPTS)/toolchain.rcpt
 		systemd_preset=$$( pkg-config --variable=systemdsystempresetdir systemd ); \
 	        echo "$${systemd_unit}/$${at_ver_rev_internal//./}-cachemanager.service" \
 	             >> $(DYNAMIC_SPEC)/$${group}/ldconfig.filelist; \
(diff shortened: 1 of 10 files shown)


grooverdan:
umask prior to creation will avoid a race condition compared to chmod.
<pre>
$ echo $SHELL
/bin/bash
$ umask 0022
$ touch /tmp/y
$ ls -la /tmp/y
-rw-r--r-- 1 danielgb danielgb 0 Aug 21 11:20 /tmp/y
</pre>

Thanks for making docker install not fail (#132).

rzinsly:
umask changes the file permissions for all files created after it, we only need to change for that one file.
FWIU this code is unlikely to get a race condition issue.

Issue #23, opened 2017-03-21
ck_provides is not working without the dummy DEB packages


rzinsly:
On DEB based distros the ck_provides is failing if the dummy packages are not installed, these packages are only used to pull the right AT dependency on apt-get.
They don't need to be installed and the test should pass without them.
We still need to test them when running in the compatible mode with the build system.

PR #156, opened 2017-08-24
Fix ck_provides for DEB packages

rzinsly:
This patch fixes the ck_provides FVTR test for DEB based distros and
adapt it to deal with the dummy packages (#23). It also fixes the
check_debs procedure to correctly get the needed packages.

### DIFF
--- a/fvtr/ck_provides/ck_provides.exp
+++ b/fvtr/ck_provides/ck_provides.exp
@@ -69,27 +69,32 @@ proc process_deb {package expected} {
 	set version "(= ${at_full_ver})"
 
 	if { ![info exists ::env(AT_WD)] } {
-		if { ![exec dpkg-query --status ${expected}]} {
+		if { [catch {exec dpkg-query --status ${expected}}] } {
 			printit "User didn't install package ${expected}.\
 Can't run this test."
 			return ${ENOSYS}
 		}
 
-		return ${ENOSYS}
+		# Verify the version of the dummy package.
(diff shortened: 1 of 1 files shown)


fvtr/ck_provides/ck_provides.exp
```diff
@@ -69,27 +69,29 @@ proc process_deb {package expected} {
...
-			printit "Provides for ${package} not found."
+		# Verify the version of the dummy package.
+		set dversion [string trimleft [exec dpkg-query --status \
+				${expected} | grep -E "Version"] "Version: "]
+		if { ![string equal ${dversion} ${at_full_ver}] } {
+			printit "Package ${expected} installed is not from AT \
+${at_full_ver}. Can't run this test."
+			return ${ENOSYS}
+		}
+
```
gftg85: It would be nice to see a comment here, something in the lines of:

# Parse the Pre-Depends field from the dummy package.  It must contain the actual package.

fvtr/ck_provides/ck_provides.exp
```diff
@@ -69,27 +69,29 @@ proc process_deb {package expected} {
...
+		set depends [string trimleft \
+				    [exec dpkg-query --status ${expected} \
+					 | grep -E "Pre-Depends"] \
+				    "Pre-Depends: "]
+
```
gftg85: Likewise:

# Check if the Pre-Depends field references the actual package.

rzinsly:
@gftg85 Done, thanks!

PR #160, opened 2017-08-29
Fix the fetch of Python patches

rzinsly:
This patch changes the fetch of the python's test_ssl fix by getting
the patch from the commit instead of the PR because the md5sum changes
when the PR changes.


rzinsly:
@gftg85 git gc usually doesn't delete objects referenced in other parts of a repository and this commit is referenced by all backports.
Anyway when the next version is released we will not need to apply this patch anymore and if this commit ID is removed someday I doubt we will still be carrying it.

PR #161, opened 2017-08-30
[fvtr] Run TBB parallel test

rzinsly:
We were not running the parallel test in our TBB test case, this patch
enables it.

### DIFF
--- a/fvtr/tbb/tbb.exp
+++ b/fvtr/tbb/tbb.exp
@@ -87,13 +87,13 @@ proc parallel_test {} {
 	printit "Testing parallel execution..."
 	if { $TARGET32 } {
 		printit "Testing dynamic 32 link..."
-		compile "${CXX} -m32 ${CFLAGS} $FULLPATH/mem-test.cpp \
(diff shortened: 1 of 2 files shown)


fvtr/tbb/tbb.exp
```diff
@@ -105,8 +105,8 @@ set tmp_file [exec mktemp]
 # Run memtest for all TBB versions.
 memtest
 
-# parallel_test is supported by TBB >= 4.2 only.
-if { ${major_ver} >= 4 && ${minor_ver} >= 2 } {
+# parallel_test is supported by TBB >= 4.3 only.
+if { ${major_ver} >= 4 && ${minor_ver} >= 3 } {
```
rzinsly: Prior to 4.2u4 we would need the tbb/internal/_flow_graph_or_impl.h header.

gftg85: It would be nice if you moved this comment (on github.com) to the code itself.
Other than that, the patch is OK.

rzinsly:
@gftg85 Done!

PR #169, opened 2017-09-18
Update many packages on AT next

rzinsly:
This patch updates the following packages on AT next to the lastest
revision:
Expat, Libhugetlbfs, Userspace RCU, MPC, MPFR, OpenSSL, Oprofile,
TBB, gperftools and Valgrind.
It also changes some of the sources to get code from master/trunk.


configs/next/packages/liburcu/stage_1
```diff
@@ -105,7 +109,7 @@ atcfg_configure ()
 	PATH=${at_dest}/bin:${PATH} \
 	CC="${configure_cc}" \
 	CFLAGS="-g -O2 -fpic -m${AT_BIT_SIZE}" \
-	${ATSRC_PACKAGE_WORK}/configure \
+	./configure \
```
gftg85: Most of the ```stage_X``` files currently use ```${ATSRC_PACKAGE_WORK}/configure```.  Do you have reason to make this change?  If so, would it be useful (easy) to add a comment about it?

rzinsly: There is no configure file in the sources of liburcu when getting the code from the repository, it is generated in the build directory by running the bootstrap script.

gftg85: OK.  Thanks for the info.

PR #196, opened 2017-10-07
Update binutils on AT next

Cherbot:
Bump to revision 0ee6e554c4f8


gftg85:
The failure on AT Next Ubuntu 16.04 ppc64le is not of of those fixed by pull request #178.  That pull request only fixed problems with test_pty and test_float_with_comma, whereas the failure is now on test_ssl.

gftg85:
The failure in test_ssl (python) also happens before this pull request.

tuliom:
Merged as 237200286243e8e3e724654c10d5ef55ab06b496.

Issue #148, opened 2017-08-10
Update expat AT 9.0 and 10.0


racardoso:
Update libexpat for AT 10.0 and 9.0.

**Notes:**

AT 9.0 should be updated to 2.1.1 and all CVE must be backported from 2.2.x version to avoid new symbols to be introduced.

AT 10.0 may be updated from 2.2.1 to 2.2.2 (or 2.2.3) since it doesn't introduces new symbols. libexpat people may not respect the MAJOR.MINOR.PATCH system. This will be decided on a discussion at mailing list.

Issues related: #84
There's a PR opened for AT 9.0: https://github.com/advancetoolchain/advance-toolchain/pull/124

Issue #84, opened 2017-06-06
[CVE-2016-0718] expat: Out-of-bounds heap read on crafted input causing crash


rzinsly:
It affects at least expat 2.1.0.
Per the Debian Security Tracker, it affects expat 2.2.0 too.

References:
- [NVD](https://nvd.nist.gov/vuln/detail/CVE-2016-0718)
- [Debian Security Tracker](https://security-tracker.debian.org/tracker/CVE-2016-0718)

racardoso:
@rzinsly It was already fixed on expat 2.2.0:
```
Changes
Release 2.2.0 Tue June 21 2016
        Security fixes:
            #537  CVE-2016-0718 -- Fix crash on malformed input
```

Not fixed for versions less than 2.1.1. I am updating AT 10.0 and 11.0 to 2.2.2 (next was already updated) do we need backport this to AT < 10.0?

In that case here's the patch:

http://www.openwall.com/lists/oss-security/2016/05/17/12

rzinsly:
> Not fixed for versions less than 2.1.1. I am updating AT 10.0 and 11.0 to 2.2.2 (next was already updated) do we need backport this to AT < 10.0?

@racardoso Yes, on AT 9.0.

racardoso:
@rzinsly what's better? apply the patch to AT 9.0 or update expat version to 2.2.2 on AT 9.0 too?

rzinsly:
@racardoso apply the patch.

racardoso:
Related to task #113 
Already updated to 2.2.2 (CVE-2016-0718 was already fixed on this version):
AT 10.0
AT 11.0
AT next

Needs to update:
AT. 9.0

PR #203, opened 2017-10-16
Update Expat on AT 9.0.

racardoso:
Update Expat to version 2.1.1 on AT 9.0 (Issues #148 and #84).
CVE fixes backported:

CVE-2012-6702
CVE-2016-5300
CVE-2016-4472
CVE-2016-0718
CVE-2016-9063
CVE-2017-9233

### DIFF
--- a/configs/9.0/packages/expat/sources
+++ b/configs/9.0/packages/expat/sources
@@ -20,15 +20,17 @@
 #
 
 ATSRC_PACKAGE_NAME="Expat XML Parser"
-ATSRC_PACKAGE_VER=2.1.0
-ATSRC_PACKAGE_DOCLINK="http://www.xml.com/pub/a/1999/09/expat/index.html"
+ATSRC_PACKAGE_VER=2.1.1
+ATSRC_PACKAGE_REV=ee0adc0d9d245
+ATSRC_PACKAGE_DOCLINK="https://libexpat.github.io/doc/"
 ATSRC_PACKAGE_RELFIXES=
-ATSRC_PACKAGE_STR_VER="${ATSRC_PACKAGE_NAME} ${ATSRC_PACKAGE_VER}"
+ATSRC_PACKAGE_STR_VER="${ATSRC_PACKAGE_NAME} ${ATSRC_PACKAGE_VER}-${ATSRC_PACKAGE_REV}"
+ATSRC_PACKAGE_VERID="${ATSRC_PACKAGE_VER}${ATSRC_PACKAGE_REV:+-${ATSRC_PACKAGE_REV}}"
 ATSRC_PACKAGE_LICENSE="MIT License"
-ATSRC_PACKAGE_PRE="test -d expat-${ATSRC_PACKAGE_VER}"
-ATSRC_PACKAGE_CO=([0]="wget -N https://downloads.sourceforge.net/project/expat/expat/${ATSRC_PACKAGE_VER}/expat-${ATSRC_PACKAGE_VER}.tar.gz")
-ATSRC_PACKAGE_POST="tar -xzf expat-${ATSRC_PACKAGE_VER}.tar.gz"
-ATSRC_PACKAGE_SRC=${AT_BASE}/sources/expat-${ATSRC_PACKAGE_VER}
+ATSRC_PACKAGE_PRE="test -d expat-${ATSRC_PACKAGE_VER}-${ATSRC_PACKAGE_REV}"
+ATSRC_PACKAGE_CO=([0]="wget -O expat-${ATSRC_PACKAGE_VERID}.tar.gz https://github.com/libexpat/libexpat/archive/${ATSRC_PACKAGE_REV}.tar.gz")
+ATSRC_PACKAGE_POST="tar xzf expat-${ATSRC_PACKAGE_VERID}.tar.gz --wildcards --strip=2 libexpat-${ATSRC_PACKAGE_REV}[^\\/]*/expat --transform=s:libexpat-${ATSRC_PACKAGE_REV}[^\\/]*/expat:libexpat-${ATSRC_PACKAGE_REV}[^\\/]*/expat-${ATSRC_PACKAGE_VERID}:"
+ATSRC_PACKAGE_SRC=${AT_BASE}/sources/expat-${ATSRC_PACKAGE_VER}-${ATSRC_PACKAGE_REV}
 ATSRC_PACKAGE_WORK=${AT_WORK_PATH}/expat
 ATSRC_PACKAGE_MLS=
 ATSRC_PACKAGE_PATCHES=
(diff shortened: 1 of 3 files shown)


racardoso:
One small note: I've changed the symlink of stage_1 on AT 10.0 to point to `8.0/.../expat/stage_1` instead. The reason why I've done that is because I still did not updated expat on AT 10.0 and it's still rely in the old stage_1 config file . After update expat on AT 10.0 and release AT 9.0 I will change it (back) to:

`9.0/../stage_1<--10.0/../stage_1<--11.0/../stage_1<--next/../stage_1`

configs/9.0/packages/expat/sources
```diff
@@ -40,12 +42,45 @@ ATSRC_PACKAGE_BUNDLE=main_toolchain
 
 atsrc_get_patches ()
 {
-        at_get_patch \
-	https://hg.mozilla.org/releases/mozilla-esr31/raw-diff/2f3e78643f5c/parser/expat/lib/xmlparse.c \
-	978566c4144504d170869619661fbd7d || return ${?}
```
gftg85: OK.  This was incorporated into expat by commit ID ba0f9c3b40c2 (between 2.1.0 and 2.1.1).

configs/9.0/packages/expat/sources
```diff
@@ -40,12 +42,45 @@ ATSRC_PACKAGE_BUNDLE=main_toolchain
...
+	# Delete Makefile that generates man
+	at_get_patch \
+		https://github.com/racardoso/libexpat/commit/a6bb3358ddae.patch \
+		bc61e4806ae531941b652bb56abb3249 || return ${?}
+
+	# Added man pages
+	at_get_patch \
+		https://github.com/racardoso/libexpat/commit/f1a8200593de.patch \
+		0c71eb2bb1b6750a96c3c5181f67de3e || return ${?}
```
gftg85: The first two patches are OK.

configs/9.0/packages/expat/sources
```diff
@@ -40,12 +42,45 @@ ATSRC_PACKAGE_BUNDLE=main_toolchain
...
+
+	#Fix CVE-2012-6702 and CVE-2016-5300
+	at_get_patch \
+		https://github.com/racardoso/libexpat/commit/95dbdf173f19.patch \
+		ea1a061d504d47aa726d09dc639db1fa || return ${?}
```
gftg85: On Debian, this patch is supposed to be applied on top of 2.1.0.  Do you know if this has the desired effect on 2.1.1?

configs/9.0/packages/expat/sources
```diff
@@ -40,12 +42,45 @@ ATSRC_PACKAGE_BUNDLE=main_toolchain
...
+
+	#Fix CVE-2016-4472
+	at_get_patch \
+		https://github.com/racardoso/libexpat/commit/ca78e04e3e67.patch \
+		413e2b92eb8b6b35b8e22a1f56a2421e || return ${?}
```
gftg85: OK.

configs/9.0/packages/expat/sources
```diff
@@ -40,12 +42,45 @@ ATSRC_PACKAGE_BUNDLE=main_toolchain
...
+
+	#Fix CVE-2016-0718
+	at_get_patch \
+		https://github.com/racardoso/libexpat/commit/be940daed613.patch \
+		1bd7c8c0414bd0a6f13acae171118adf || return ${?}
```
gftg85: This looks wrong.

The original patch can be found on [centos](https://git.centos.org/raw/rpms/expat/c7/SOURCES/expat-2.1.0-CVE-2016-0718.patch).  However, the original has a hunk that your patch does not have:

```
@@ -1744,7 +1744,7 @@
                        const char *end,
                        POSITION *pos)
 {
-  while (ptr < end) {
+  while (end - ptr >= MINBPC(enc)) {
     switch (BYTE_TYPE(enc, ptr)) {
 #define LEAD_CASE(n) \
     case BT_LEAD ## n: \
```

Since all the files modified by the original patch are in the same subdirectory (expat), you could use `-d`, instead of providing your own:

```
patch -p1 -d expat < expat-2.1.0-CVE-2016-0718.patch
```

configs/9.0/packages/expat/sources
```diff
@@ -40,12 +42,45 @@ ATSRC_PACKAGE_BUNDLE=main_toolchain
...
+
+	#Fix CVE-2016-9063 and CVE-2017-9233
+	at_get_patch \
+		https://github.com/racardoso/libexpat/commit/a39744e4ec1e.patch \
+		50c236fd76f8868a48468faff968fc9f || return ${?}
```
gftg85: The commit IDs listed in this patch are from libexpat itself and they apply without errors, so wouldn't it be better to just download them and use them directly:

```
at_get_patch \
  https://github.com/libexpat/libexpat/commit/d4f735b88d9932bd5039df2335eefdd0723dbe20.patch
  [...]
at_get_patch \
  https://github.com/libexpat/libexpat/commit/c4bf96bb51dd2a1b0e185374362ee136fe2c9d7f.patch
  [...]

[...]
patch -p1 < d4f735b88d9932bd5039df2335eefdd0723dbe20.patch
patch -p1 < c4bf96bb51dd2a1b0e185374362ee136fe2c9d7f.patch
```

configs/9.0/packages/expat/sources
```diff
@@ -40,12 +42,45 @@ ATSRC_PACKAGE_BUNDLE=main_toolchain
...
 }
```
gftg85: Do you plan to maintain your clone of libexpat available forever, and do you plan to keep the branches where these patches belong to, i.e.: branch [backport_R_2_1_1](https://github.com/racardoso/libexpat/commits/backport_R_2_1_1), also forever?

Otherwise, it would be better to move it somewhere else, e.g.: to [powertechpreview](https://github.com/powertechpreview/powertechpreview/tree/master/Expat%20Patches).

configs/9.0/packages/expat/sources
```diff
@@ -40,12 +42,45 @@ ATSRC_PACKAGE_BUNDLE=main_toolchain
...
 
 atsrc_apply_patches ()
 {
-	patch -p3 < xmlparse.c || return ${?}
+	patch -p2 < a6bb3358ddae.patch || return ${?}
+	patch -p2 < f1a8200593de.patch || return ${?}
+	patch -p2 < 95dbdf173f19.patch || return ${?}
+	patch -p2 < ca78e04e3e67.patch || return ${?}
+	patch -p2 < be940daed613.patch || return ${?}
+	patch -p2 < a39744e4ec1e.patch || return ${?}
```
gftg85: All these patches seem to be generated with git from the root directory, so I would expect this to be `-p1`, not `-p2`.  Didn't you get any errors?

gftg85:
I tried to add more inline comments to the code, but I'm getting an error from github ("Position is invalid").  So I'll send the comments here:

In configs/9.0/packages/expat/stage_1, line 25:
> +ATCFG_BUILD_STAGE_T='link'

I do not oppose this change as it is, but it would have been nicer if an explanation about the change from `dir` to `link` had been provided.

It would also have been nicer if the change from a symlink (to configs/9.0/packages/expat/stage_1) to a regular file with the same contents had been provided (such as commit ID 68114b1), and only then the content change in a separate patch.

In configs/9.0/packages/expat/stage_1, line 31:
>+atcfg_configure() {

Likewise, an explanation about the need for this new `atcfg_pre_configure` step would have been nice. I understand that this change is the same from commit ID 488800c, but that commit ID doesn't provide an explanation, either.

configs/9.0/packages/expat/sources
racardoso: @gftg85 Yes it has. Those fixes are related to the poor entropy that was used on libexpat. Those fixes was available upstream only to versions 2.2.0 or higher.

configs/9.0/packages/expat/sources
racardoso: @gftg85 That's odd. I'm not sure why centos people includes this diff on their file. I get this patch from [Debian](http://www.openwall.com/lists/oss-security/2016/05/17/12) and there's no such diff.

configs/9.0/packages/expat/sources
racardoso: @gftg85 See the answer bellow.

configs/9.0/packages/expat/sources
racardoso: @gftg85 Yes. I am not planing to maintain this branch forever. In fact I will have to migrate those patches to powertechpreview I will do that.

configs/9.0/packages/expat/sources
gftg85: OK.  I haven't seen this patch from Debian.  I only followed the link to patchwork, and then I found the centos patch.

configs/9.0/packages/expat/sources
racardoso: @gftg85 In fact I would expect it too. I used `-p1` when apply those patches on expat source here. But on advance toolchain it can't apply the patch with `-p1`. I think advance toolchain scripts copies the patch to the source folder and then apply the patch. Since the patch is something like `a/expat/lib/xmlparse.c b/expat/lib/xmlparse.c` I guess I should have to use `-p2` instead.

racardoso:
@gftg85 

> In configs/9.0/packages/expat/stage_1, line 25:

>     +ATCFG_BUILD_STAGE_T='link'

> I do not oppose this change as it is, but it would have been nicer if an explanation about the change from dir to link had been provided.

The problem here is because expat tar.gz releases from github doesn't include configure.ac. It must be generated by autotools. When I call autotools (aclocal, automake, autoreconf) it generates configure.ac on a source/expat folder using `at_pre_configure` but try to build it on another folder with `at_configure` and can't find this file. So we changed `dir` to `link` in order to build expat in the same folder. In fact we don't need to change the folder here anyway.

> It would also have been nicer if the change from a symlink...

I can change that. I was do that on on the same commit because if not we have one commit the will not build and I am on kind of rush to fix this problem in order to release AT 9.0.

> In configs/9.0/packages/expat/stage_1, line 31:

>    +atcfg_configure() {

Like I say before, different from sourcecode versions, tar.gz releases from github doesn't include configure.ac. It must be generated by autotools. So I call pre-configure to call autotools and generate configure.ac.

gftg85:
Thanks for the new patch.

It looks good to me, but I suggest squashing the two patches into a single commit before pushing to the repository.  I don't see any advantage in keeping them separate.

racardoso:
@gftg85 I will rebase before merge here. I only sent another commit to make easy to review.

racardoso:
@tuliom There's anything you want to comment before I merge this PR?

PR #213, opened 2017-10-23
Prepare for AT 9.0-1

racardoso:
Signed-off-by: Rogerio Alves <rcardoso@linux.vnet.ibm>


tuliom:
Actually, 9.0-6 RC1.

racardoso:
@gftg85 @tuliom Yes my mistake. But I have another question before send the fix. Should it be 9.0-6 RC1 or 9.0-7 RC1?

gftg85:
> Should it be 9.0-6 RC1 or 9.0-7 RC1?

9.0-6 RC1

When commit ID 6db2a02f0e520b3bb514938d6b539450f6074b52 opened AT 9.0 for development, it already set the correct version number.  The previous release was AT 9.0-5 (unfortunately, this information is not available in the repository, because it predates the migration to github.com).

PR #220, opened 2017-10-27
Fix python test case failures on AT next

gftg85:
The test case, test_ssl, in python was failing two subtests: test_multiprocessing_forkserver and
test_random_fork.  This pull requests fixes them.

Tested for SLES 12, Ubuntu 14.04, and Ubuntu 16.04 (Should fix the CI failures).

### DIFF
--- a/configs/11.0/packages/python/sources
+++ b/configs/11.0/packages/python/sources
@@ -49,12 +49,18 @@ atsrc_get_patches ()
 	at_get_patch \
 		http://pkgs.fedoraproject.org/cgit/rpms/python3.git/plain/00273-skip-float-test.patch \
 		a1cb79875e6c2633222dc95910e59ab8 || return ${?}
+
+	# Fix for test_ssl
+	at_get_patch \
+		https://raw.githubusercontent.com/powertechpreview/powertechpreview/93548fdd0ae7740f7a6158ced0ebdbd5a125bacf/Python%20Fixes/python-3.6.3-reseed_openssl_after_fork.patch \
+		2d8bd979783852a291ad5b916f845202 || return ${?}
 }
 
 atsrc_apply_patches ()
 {
 	patch -p1 < python-3.6.0-getlib64s1.patch || return ${?}
 	patch -p1 < 00273-skip-float-test.patch || return ${?}
+	patch -p1 < python-3.6.3-reseed_openssl_after_fork.patch || return ${?}
 }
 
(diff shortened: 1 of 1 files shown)


gftg85:
The bug report I found is https://bugs.python.org/issue18747#msg195774, which contains patches for some series (this one is for the 3.3 series).

However, this is not how I actually found it.
I found it searching for `test_ssl self.assertNotEqual(child_random, parent_random)` on the web, which brought these results: [link1](http://grokbase.com/t/python/python-checkins/138n5my3xy/cpython-3-3-issue-18747-re-seed-openssls-pseudo-random-number-generator-after-fork) and [link2](https://github.com/google/cpython-pt/commit/a90ba25e7aaf1d03381f51dd05ce6947529f8d23).

I removed the NEWS entry from them because it did not apply.  But maybe I should change the patch so that it applies.  The NEWS entry contains the information about the bug report.

RajalakshmiS:
LGTM

tuliom:
@gftg85, thank you for clarifying.

According to [comment #200416](https://bugs.python.org/issue18747#msg200416) from that issue, this patch seems to cause serious errors.
So, patch #1 is fine.  Patch #2 doesn't seem a good idea...

gftg85:
Indeed, and it's been reverted by https://github.com/python/cpython/commit/fb6b44e83083df34e8fccced4a2b518b27718144.

Thanks for the review.

tuliom:
@gftg85, but I think we have enough reason to disable the test that's failing...

gftg85:
The algorithm for pseudo random number generation on openssl suffered various changes in the development branch (master; target 1.1.1-dev).  Particularly, the following commits:

https://github.com/openssl/openssl/commit/2915fe19a6676374c335d8c50eaaa4c940cf47d6
https://github.com/openssl/openssl/commit/a35f607c9f9112c649b367d05639394fc1c30771

With some explanations about use in 

https://github.com/openssl/openssl/blob/master/doc/man3/OPENSSL_fork_prepare.pod


I wonder if something has to be done in python to adapt to this new behavior.  I'm not completely sure that we should disable the test (test_random_fork).

tuliom:
@gftg85, that may be the case, but according to the documentation you pointed:

> Platforms such as Linux that have both functions will normally not need to call these functions as the OpenSSL library will do so automatically.

Maybe Python has something special?

PR #221, opened 2017-10-31
Fix version info about glibc on AT next

tuliom:
Since glibc 2.26, the community started to tag development branches with
revision .9000, e.g. 2.26.9000.  This patch adheres to the same
nomenclature.
It also removes "-ibm" from the version string as this is coming
directly from master instead of one of the ibm/ branches.


RajalakshmiS:
LGTM

tuliom:
Merged.

PR #223, opened 2017-10-31
remote-build: improve error check when cleaning files

tuliom:
On some distros, the check for the availability of remake.sh was
returning false even if the file was there, causing unexpected build
behavior.

This patch uses the shell from the remote system to validate if
remake.sh is available and run remake.sh clean without requiring extra
expect code.


RajalakshmiS:
LGTM

tuliom:
Merged.

Issue #204, opened 2017-10-20
libstdc++ for POWER9 doesn't have std::__once_callable


tuliom:
When building libstdc++ for POWER9, the configure sets gcc_cv_have_tls=no instead of gcc_cv_have_tls=yes.
This is likely not setting _GLIBCXX_HAVE_TLS and removes std::__once_callable.

Why is gcc_cv_have_tls set to "no"?

tuliom:
It affects AT 10.0 and 11.0.

tuliom:
The issue has been fixed and will be available in the next releases of AT 10.0 and 11.0.

PR #222, opened 2017-10-31
fvtr: Ignore ck_tuned on AT <= 6.0

tuliom:
AT <= 6.0 didn't use  to provide tuned libraries.

Signed-off-by: Tulio Magno Quites Machado Filho <tuliom@linux.vnet.ibm.com>

PR #224, opened 2017-10-31
Fix issue 204

tuliom:


tuliom:
AT 7.0 RHEL6 ppc64 85 failed due to disk space issues.
I retried that build as AT 7.0 RHEL6 ppc64 86 and it passed.

AT 6.0 RHEL6 ppc64 is being fixed in PR #222.
AT next SLES12 ppc64le and AT next Ubuntu16 ppc64le failed in the python testcases.

configs/11.0/packages/gcc/stage_optimized
```diff
@@ -93,10 +93,15 @@ atcfg_make() {
...
+	# ssp_have_usable_vsnprintf=define is required to avoid configure
+	# scripts from running programs compiled with a -mcpu that is superior
+	# of the base value, which can cause a silent illegal instruction
+	# error, forcing the script to assume that vsnprintf is not usable.
 	PATH=${at_dest}/bin:${PATH} \
 	libgo_cv_lib_setcontext_clobbers_tls=no \
 		${SUB_MAKE} all \
-			    gcc_cv_have_tls=yes
+			    gcc_cv_have_tls=yes \
+			    ssp_have_usable_vsnprintf=define
```
gftg85: It's fine to keep these variables here, but is there a reason to not keep them together with the other cached variable (libgo_cv_lib_setcontext_clobbers_tls)?

tuliom: AFAIK, variables defined as env vars are not passed to subshells unless they're explicitly exported or if the underlying shell receives them.
I have already confirmed that ssp_have_usable_vsnprintf is affected by this.

tuliom:
Merged.
This fixes issue #204 .

PR #244, opened 2017-11-18
Update glibc on AT next

Cherbot:
Bump to revision df0c40ee3a89


gftg85:
I suspected that the failure on Ubuntu 14 (build 486) was not related to the update, so I asked for a rebuild and it passed, as I expected.  Approving.

PR #257, opened 2017-11-24
Adapt update_revision.sh to get code from branches

rzinsly:
This patch changes the update_revision script to be able to get the latest
revision from the current branch used in the sources file instead of the
latest revision from the repository.

### DIFF
--- a/scripts/utilities/update_revision.sh
+++ b/scripts/utilities/update_revision.sh
@@ -128,14 +128,32 @@ get_latest_revision ()
 		return 1;
 	fi
 
-	local isGit=$(echo ${1} | grep -ce "git:" -e "\.git$");
+	local url=${1}
+	local isGit=$(echo ${url} | grep -ce "git:" -e "\.git$")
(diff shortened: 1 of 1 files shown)


scripts/utilities/update_revision.sh
```diff
@@ -128,14 +128,32 @@ get_latest_revision ()
...
 	local hash="";
 
 	if [[ "${isGit}" == 1 ]]; then
-		hash=$(git ls-remote ${1} HEAD | cut -f1)
+		local tmp_dir=$(mktemp -d)
+		git clone ${url} ${tmp_dir}
+		pushd ${tmp_dir} > /dev/null
+		# Get which branch is being used.
+		local branch=$(git branch -r --contains ${ATSRC_PACKAGE_REV} \
+			| cut -d/ -f2-)
```
rzinsly: One alternative is to start carrying the branch in a variable in the sources file, IMO I prefer this patch's approach.

PR #263, opened 2017-12-06
Update many packages on AT 11.0 and next

rzinsly:
These patches change the source file for many packages to be able to get the code from stable or development branches.
This will help us to automatic update more packages on AT 11.0 and next.


configs/11.0/packages/openssl/sources
```diff
@@ -20,15 +20,17 @@
 #
 
 ATSRC_PACKAGE_NAME="Open Secure Sockets Layer"
-ATSRC_PACKAGE_VER="1.1.0g"
+ATSRC_PACKAGE_VER="1.1.0"
```
tuliom: Isn't better to keep it 1.1.0g in order to provide the release info?

configs/11.0/packages/python/sources
```diff
@@ -20,15 +20,17 @@
 #
 
 ATSRC_PACKAGE_NAME="Python"
-ATSRC_PACKAGE_VER=3.6.3
+ATSRC_PACKAGE_VER=3.6
```
gftg85: The same comment made with respect to openssl applies to this file, as well.  However, some packages on advance-toolchain have this kind of point-release information, while others don't.  I'm OK with this change, and I would also be OK without it.

RajalakshmiS:
LGTM

PR #284, opened 2017-12-23
Update glibc on AT next

Cherbot:
Bump to revision ed8ae46bed42


gftg85:
This fails because the local patch [1] to enable lock elision no longer applies.  It doesn't apply any longer, because elision enabling/disabling is integrated into glibc master [2].

The local patch needs to be removed from AT Next.

[1] https://raw.githubusercontent.com/powertechpreview/powertechpreview/26747dc0bfb0f8f30de9ebad78847a96d3191ba5/GLIBC%20PowerPC%20Backport/2.22/0001-powerpc-Conditionally-enable-TLE.patch

[2] https://sourceware.org/git/?p=glibc.git;a=commit;h=07ed18d26a342741cb25a4739158c65ed9dd4d09

gftg85:
My mistake.

The failure is not because of the elision patch, but because of our local patch for RUNPATH/RPATH [1].

Since commit ID 9d7a3741c9e59eba87fb3ca6b9f979befce07826 on glibc [2], the code in elf/get-dynamic-info.h changed, and our local patch no longer applies.

Should we just update the local patch so that it applies on current master, or has this patch become obsolete?

[1] https://raw.githubusercontent.com/powertechpreview/powertechpreview/2b67d9b6b6b5f8914e8579ad50ec2c69ddcf6198/GLIBC%20PowerPC%20Backport/2.26/0001-Remove-assert-if-DT_RUNPATH-and-DT_RPATH-flags-are-f.patch

[2] https://sourceware.org/git/?p=glibc.git;a=commit;h=9d7a3741c9e59eba87fb3ca6b9f979befce07826

tuliom:
> Should we just update the local patch so that it applies on current master, or has this patch become obsolete?

We need to update it, because both asserts are still being used.

gftg85:
This is now being addressed by #296

rzinsly:
#296 was merged.

PR #290, opened 2017-12-23
Update openssl on AT 9.0

Cherbot:
Bump to revision 71d53e8ba5b9


configs/9.0/packages/openssl/sources
```diff
@@ -21,7 +21,7 @@
 
 ATSRC_PACKAGE_NAME="Open Secure Sockets Layer"
 ATSRC_PACKAGE_VER="1.0.2n"
```
gftg85: This version is still valid (no new tags have been created).

PR #292, opened 2017-12-30
Update mpfr on AT next

Cherbot:
Bump to revision 12056


configs/next/packages/mpfr/sources
```diff
@@ -21,7 +21,7 @@
 
 ATSRC_PACKAGE_NAME="GNU MPFR Library"
 ATSRC_PACKAGE_VER=3.1.7
```
gftg85: I'm not opposed to this automatic update, but I think we should update `ATSRC_PACKAGE_VER` to `4.1.0-dev`, because that's the version reported by:

```
$ svn cat https://scm.gforge.inria.fr/anonscm/svn/mpfr/trunk/VERSION
4.1.0-dev
```

gftg85: > I'm not opposed to this automatic update

This means that I believe this patch should be committed with or without the change to `ATSRC_PACKAGE_VER `.  If changing the commit is possible from within the pull request, I suggest this be done.

configs/next/packages/mpfr/sources
```diff
@@ -20,8 +20,8 @@
 #
 
 ATSRC_PACKAGE_NAME="GNU MPFR Library"
-ATSRC_PACKAGE_VER=3.1.7
-ATSRC_PACKAGE_REV=11775
+ATSRC_PACKAGE_VER=4.1
```
rzinsly: @gftg85 I prefer to use 4.1 instead of 4.1.0-dev because as we are following the master branch the point release can change often.

configs/next/packages/mpfr/sources
```diff
@@ -20,8 +20,8 @@
...
+ATSRC_PACKAGE_REV=11945
```
gftg85: The original pull request was updating to revision 12056.   Why is this updating to an older version, now?

rzinsly: Oops, not sure what happened.
Fixed now.

PR #330, opened 2018-01-25
[fvtr] Add SPEC CPU2017 to our FVTR

rzinsly:
This patch adds the SPEC CPU2017 testsuite to our FVTR.

### DIFF
--- a/fvtr/speccpu2006/speccpu2006.exp
+++ b/fvtr/speccpu2006/speccpu2006.exp
@@ -19,8 +19,9 @@
 
 source ./shared.exp
 
-if { [array names env -exact SPEC] == "" } {
+if { [array names env -exact SPEC] == "" || ! [regexp "cpu2006" $env(SPEC)] } {
 	printit "Skipping: Didn't find an installed SPEC CPU 2006."
+	printit "The directory needs to have \"cpu2006\" in its name ."
 	printit "Check if it's environment is loaded."
 	exit $ENOSYS
 }
--- a/fvtr/speccpu2017/at-sniff32.cfg
+++ b/fvtr/speccpu2017/at-sniff32.cfg
@@ -1,0 +1,73 @@
+#  Copyright 2018 IBM Corporation
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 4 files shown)


fvtr/speccpu2017/speccpu2017.exp
```diff
@@ -0,0 +1,124 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Check if AT can compile and run some SPEC CPU 2017 benchmarks
+# The SPEC CPU 2017 must be installed and it's environment set, as explained at
+# http://www.spec.org/cpu2017/Docs/install-guide-unix.html#s6
+
+source ./shared.exp
+
+if { [array names env -exact SPEC] == "" } {
```
gftg85: How do you distinguish that SPEC CPU**2017** is installed (and active (i.e.: `shrc` was executed)) rather than SPEC CPU**2006**?  If I recall correctly, both set the `SPEC` environment variable.

rzinsly: @gftg85: You are correct.
The only way I can think right now is to check the name of the directory in the SPEC environment variable, but this way we need to force spec to be installed with a given name.
I modified the patch to check for cpu2017 or cpu2006.
Do you have any suggestions?

fvtr/speccpu2006/speccpu2006.exp
```diff
@@ -19,8 +19,9 @@
 
 source ./shared.exp
 
-if { [array names env -exact SPEC] == "" } {
+if { ! [regexp "cpu2006" $env(SPEC)] } {
 	printit "Skipping: Didn't find an installed SPEC CPU 2006."
+	printit "The directory needs to have \"cpu2006\" in its name ."
```
gftg85: This change is OK, but I believe it would be better to change the commit message so that it reflects the fact that the patch affects SPEC CPU2006 as well.  Currently, the commit message gives the impression that only SPEC CPU2017 is affected.

Updating the commit message will help people understand why their environment is no longer running SPEC (if said people did not use cpu2006 in the directory name).

fvtr/speccpu2017/at-sniff32.cfg
```diff
@@ -0,0 +1,74 @@
...
+
+# How to say "Show me your version, please"
+CC_VERSION_OPTION       = -v
+CXX_VERSION_OPTION      = -v
+FC_VERSION_OPTION       = -v
+
+
+#--------- Portability --------------------------------------------------------
+default:               # data model applies to all benchmarks
+EXTRA_PORTABILITY = -DSPEC_LP64
```
tuliom: LP64 is usually related to 64-bit machines.
What is `SPEC_LP64` doing in this case for 32-bit machines.

PR #337, opened 2018-01-27
Update libpfm on AT next

Cherbot:
Bump to revision 18e3c1f0254a


configs/next/packages/libpfm/sources
```diff
@@ -21,7 +21,7 @@
 
 ATSRC_PACKAGE_NAME="Libpfm"
 ATSRC_PACKAGE_VER=4.8.0
```
tuliom: Since commit ID e4791b that libpfm version should have been updated to 4.9.0.

PR #340, opened 2018-01-27
Update gdb on AT 11.0

Cherbot:
Bump to revision 621a88c13ad3


configs/11.0/packages/gdb/sources
```diff
@@ -20,7 +20,7 @@
 #
 ATSRC_PACKAGE_NAME="GDB (GNU Debugger)"
 ATSRC_PACKAGE_VER=8.0
```
tuliom: Should we update this to 8.0.1?

rzinsly:
All Ubuntu 16.04 builds is failing on tcbot due to the oprofile test.

Issue #179, opened 2017-09-28
Cache manager service isn't properly removed


tuliom:
I've noticed that AT 11 is leaving some cache manager files on Ubuntu 16.04 after being uninstalled, e.g.:

/etc/systemd/system/multi-user.target.wants/at11.0-0-beta1-cachemanager.service -> /lib/systemd/system/at11.0-0-beta1-cachemanager.service

rzinsly:
Fixed by #345.

PR #345, opened 2018-02-01
Properly remove the cache manager services on Ubuntu

rzinsly:
Removes the systemd symlinks for our cache manager system after uninstallation on Ubuntu (#179).


rzinsly:
The Ubuntu builds failed due to #354.

rzinsly:
@tuliom Done!

PR #350, opened 2018-02-03
Update glibc on AT 11.0

Cherbot:
Bump to revision 400747ec4ff4


rzinsly:
The Ubuntu build failed due to #354.

PR #351, opened 2018-02-03
Update gcc on AT 10.0

Cherbot:
Bump to revision 257243


rzinsly:
The Sles 12 and Ubuntu 16.04 builds failed due to #354.

PR #355, opened 2018-02-06
Fix environment modules symlinking

### LINKED ISSUE #318: environment module installation to system directory is broken
On RHEL 7:
$ readlink /usr/share/modules/modulefiles/at11.0
/share/modules/modulefiles/at11.0

The link should point to /opt/at11.0/share/modules/modulefiles/at11.0.

ThinkOpenly:
Environment modules are shipped in both cross and non-cross packages, then symlinked in to the system environment modules directory for easy use.  The cross packages were not correctly allowing for installation relocation (via "rpm --prefix"), and the non-cross packages were unnecessarily using RPM_INSTALL_PREFIX (even though they are not relocatable.

Fixes #318 

Signed-off-by:  Paul A. Clarke <pc@us.ibm.com>

### DIFF
--- a/configs/11.0/specs/monolithic_cross.spec
+++ b/configs/11.0/specs/monolithic_cross.spec
@@ -112,12 +112,13 @@ gzip -9nvf ${RPM_BUILD_ROOT}%{_tgtinfodir}/*.info*
 for INFO in $(ls ${RPM_INSTALL_PREFIX}/%{tgtinfodir_r}/*.info.gz); do
 	install-info ${INFO} ${RPM_INSTALL_PREFIX}/%{tgtinfodir_r}/dir
 done
+datadir="${RPM_INSTALL_PREFIX}/%{datadir_r}"
 if [[ -w /usr/share/modules/modulefiles/. ]]; then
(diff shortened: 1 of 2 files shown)


configs/11.0/specs/monolithic_cross.spec
```diff
@@ -113,11 +113,11 @@ for INFO in $(ls ${RPM_INSTALL_PREFIX}/%{tgtinfodir_r}/*.info.gz); do
 	install-info ${INFO} ${RPM_INSTALL_PREFIX}/%{tgtinfodir_r}/dir
 done
 if [[ -w /usr/share/modules/modulefiles/. ]]; then
-	ln -sf %{_datadir}/modules/modulefiles/%{at_dir_name}-%{target} \
+  ln -sf ${RPM_INSTALL_PREFIX}/%{datadir_r}/modules/modulefiles/%{at_dir_name}-%{target} \
```
rzinsly: The tab was replaced by two spaces and the line still passes the 80th column.
You could create a variable with "${RPM_INSTALL_PREFIX}/%{datadir_r}" and keep the tab and the column limit.

PR #363, opened 2018-02-20
fvtr: Remove Oprofile and Python temporary files after testing

### LINKED ISSUE #354: [fvtr] Oprofile and Python tests leave files behind
After running the Oprofile FVTR test the oprofile_data directory and its files is left at fvtr/ directory.
After running the Python tests some tmp files are left in the test directory.

This could affect the next tests as this is not removed even after a cleanall.

rzinsly:
The Oprofile and Python FVTR testsuites were leaving test files after
the tests runs, this patch ensure they are removed.
This fixes #354.

### DIFF
--- a/fvtr/oprofile/oprofile.exp
+++ b/fvtr/oprofile/oprofile.exp
@@ -129,6 +129,8 @@ if { ${TARGET64} } {
 	lappend wordsize_l 64
 }
 
+cd ${FULLPATH}
+
 set rc [expr {$rc || [test_this "ophelp" ${FULLPATH}/ophelp.log]}]
 
 set test_source "${FULLPATH}/test.c"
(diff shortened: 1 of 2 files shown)


fvtr/python/python.exp
```diff
@@ -121,21 +121,25 @@ expect {
 	# Wait for 1 hour at most
 	-timeout 3600
 	-re "(test\[s\]* failed|Test Failed)" {
-		fvtr_error "Failed to run Python tests."
+		printit "Failed to run Python tests."
```
tuliom: Isn't it missing `$ERROR`? e.g.

```
printit "Failed to run Python tests." $ERROR
```

fvtr/oprofile/oprofile.exp
```diff
@@ -129,6 +129,8 @@ if { ${TARGET64} } {
 	lappend wordsize_l 64
 }
 
+cd ${FULLPATH}
```
tuliom: `${FULLPATH}` usually has the value returned from pwd with a '/' in appended.
Could you elaborate how is this command helping to fix the bug, please?

tuliom: Oh! I understand what's happening now and how this is helping.
Have we changed the pwd of the tests?  Interesting...

rzinsly: @tuliom Not that I'm aware of.

PR #370, opened 2018-02-24
Update mpfr on AT next

Cherbot:
Bump to revision 12438


tuliom:
Build 603 on AT next SLES12 failed due to a disk error.
I restarted the build.

gftg85:
Even though the rebuilds succeeded, I think we should just discard this pull request, as many other packages, including gcc, have been update recently.

PR #381, opened 2018-03-06
Update the URL of a patch to Python.

tuliom:
The patch was removed from master and the old URL is not available
anymore.
Replace it with an URL that carries the commit ID in order to avoid this
kind of problem.


rzinsly:
I didn't notice the python on AT next is not symlinked to AT 11.0, rather than updating the URL I prefer to update python to the same version as AT 11.0 that doesn't need that patch anymore.

tuliom:
I've replaced the AT next file with a symlink, but I'm closing this PR without merging it.
The new version of the patch is available in PR #385, which can't be tested manually without this patch.
So, in order to avoid duplicating this patch in different PRs, I'm closing this one.

PR #375, opened 2018-03-03
Update gcc on AT next

Cherbot:
Bump to revision d71fa3718231

PR #385, opened 2018-03-07
Update GCC on AT next

tuliom:
The first patches prepare other packages to build with the latest revision from GCC 8.

The last patch is the final update.

I think PR #375 should be discarded now.


tuliom:
Notice the first patch is being reviewed as PR #381.

tuliom:
It seems GCC didn't like my changes to libpfm on AT 6.0.  I'm going to recreate this patch and remove that symlink.

tuliom:
I updated this patchset to version 2.
I hope that PR #381 can be reviewed here now.

rzinsly:
Not sure why but seems our CI system tested commit 2eae851 instead of 1432159.

tuliom:
2eae851776cc32c086cce2876d07de03afab98e1 is indeed the right commit ID to be tested.  Github doesn't exhibit the right commit log.

tuliom:
I've disabled Python's test_random_fork based on aed7470235e2b49be1ceeff621996ad716e5fc73.
This should solve the Python failures in the previous version.

tuliom:
Merged.

PR #386, opened 2018-03-10
Update binutils on AT next

Cherbot:
Bump to revision 7d4f13635fd2


configs/next/packages/binutils/sources
```diff
@@ -21,7 +21,7 @@
 
 ATSRC_PACKAGE_NAME="GNU Binutils"
 ATSRC_PACKAGE_VER=2.29.90
-ATSRC_PACKAGE_REV=d4333bab0a09
+ATSRC_PACKAGE_REV=7d4f13635fd2
```
gftg85: The version should have been bumped to 2.30.51, since binutils commit ID 769c7ea50720

PR #388, opened 2018-03-10
Update mpc on AT next

Cherbot:
Bump to revision c550190e067a


configs/next/packages/mpc/sources
```diff
@@ -21,7 +21,7 @@
 
 ATSRC_PACKAGE_NAME="GNU MPC Library"
 ATSRC_PACKAGE_VER=1.0.4
-ATSRC_PACKAGE_REV=fc7e1fcb28eb
+ATSRC_PACKAGE_REV=c550190e067a
```
gftg85: This revision update requires an update in the package version, from 1.0.4 to 1.1.0.

PR #398, opened 2018-03-20
Fix locale generation

tuliom:
The first patch limits the number of jobs running in parallel when generating locales.

The second patch guarantees a build failure in case the locale generation fails.


tuliom:
Merged

PR #403, opened 2018-03-24
Update mpfr on AT next

Cherbot:
Bump to revision 12498


configs/next/packages/mpfr/sources
```diff
@@ -21,7 +21,7 @@
 
 ATSRC_PACKAGE_NAME="GNU MPFR Library"
 ATSRC_PACKAGE_VER=4.1
```
gftg85: In #292, I mentioned that the `VERSION` file on mpfr states the the current development version is 4.1.0-dev:

```
$ svn cat https://scm.gforge.inria.fr/anonscm/svn/mpfr/trunk/VERSION
4.1.0-dev
```

At the time, we decided to set `ATSRC_PACKAGE_VER` to 4.1.  Now, however, I feel that it is misleading, because 4.1 has never actually been released.  We should probably set it to 4.0, which is the last stable branch, as can be seen on the following command and output.

```
$ svn ls https://scm.gforge.inria.fr/anonscm/svn/mpfr/branches
2.0.2/
2.1/
2.2/
2.3/
2.4/
3.0/
3.1/
4.0/
```

On the other hand, this is AT Next, which is never released.  So, if we don't change `ATSRC_PACKAGE_VER` to 4.0 now, we just have to be careful when cutting the branch for AT 12.0.

rzinsly: I believe we should keep this on 4.1, AT next follows the unstable branch which has the code that will become 4.1, and should be closer to that than to 4.0.
When we start working in AT 12.0 we shouldn't be following the same branch so the version will be different.

PR #441, opened 2018-04-16
[install_tools] Add option to update packages

rzinsly:
This patch changes the install tools script to add the option to update
packages.

### DIFF
--- a/utils/install_tools.sh
+++ b/utils/install_tools.sh
@@ -22,7 +22,7 @@
 
 # Install AT packages on PowerPC
 #
-# install_native <at_version>
+# install_native <at_version> <command>
 install_native ()
 {
 	local version=${1}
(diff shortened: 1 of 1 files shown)


utils/install_tools.sh
```diff
@@ -60,30 +60,119 @@ install_cross ()
...
+install_last_native ()
+{
+	local version=${1}
+
+	if [[ ${debian} == "yes" ]]; then
+		sudo aptitude install -y \
+		   advance-toolchain-at${version}-runtime \
+		   advance-toolchain-at${version}-devel \
+		   advance-toolchain-at${version}-mcore-libs \
+		   advance-toolchain-at${version}-perf \
```
tuliom: Shouldn't debuginfo packages be installed in order to test their updates too?

utils/install_tools.sh
```diff
@@ -60,37 +61,137 @@ install_cross ()
...
+{
+	local version=${1}
+
+	if [[ ${debian} == "yes" ]]; then
+		sudo apt-get install -y \
+		   advance-toolchain-at${version}-runtime \
+		   advance-toolchain-at${version}-runtime-dbg \
+		   advance-toolchain-at${version}-devel* \
+		   advance-toolchain-at${version}-mcore-libs* \
+		   advance-toolchain-at${version}-perf* \
```
tuliom: Why do you specify `-dbg` for runtime, but you using `*` for the rest of the packages?

rzinsly: Because If we use * with the runtime we will get the compat packages too, they will conflict with the main runtime.

PR #485, opened 2018-06-02
Update gcc on AT next

Cherbot:
Bump to revision b084abe1fe68


tuliom:
This is breaking Binutils build:

```
sources/binutils/gold/incremental.h:1704:9: error: implicitly-declared ‘constexpr gold::Incremental_binary::Input_reader::Input_reader(const gold::Incremental_binary::Input_reader&)’ is deprecated [-Werror=deprecated-copy]
   class Sized_input_reader : public Input_reader
         ^~~~~~~~~~~~~~~~~~
```

Alan Modra already has a [patch](https://sourceware.org/ml/binutils/2018-06/msg00015.html).  We have to wait for it to be accepted on Binutils before merging this GCC update.

tuliom:
This issue should already been fixed in Binutils with the following commit:

```
commit a1893a821c4a2e953d13b97360b85650f2127134
Author: Cary Coutant <ccoutant@gmail.com>
Date:   Fri Jun 22 18:19:51 2018 -0700

    Silence GCC 9 error about deprecated implicit copy constructor.
    
    Replacing push_back() with emplace_back() eliminates the calls to the
    copy constructor, but I still had to provide explicit copy constructors
    because of the call to vector::reserve(), which tries to instantiate them
    even though they'll never actually be called when reserve() is called
    on an empty vector.
    
    gold/
            * incremental.cc (Sized_incremental_binary::setup_readers): Use
            emplace_back for GCC 5 and later.
            * incremental.h (Incremental_binary::Input_reader): Provide copy
            constructor.
            (Sized_incremental_binary::Sized_input_reader): Likewise.
```

We need to rebase this patch to confirm.

Issue #497, opened 2018-06-18
GCC revision does not match


tuliom:
AT 11.0-3 was released with GCC rev. 257474, however, if I run `/opt/at11.0/bin/gcc -v` I get:
`gcc version 7.3.1 20180207 (Advance-Toolchain-at11.0) [ibm/gcc-7-branch revision 257393] (GCC)`

The revision information is stored in file `gcc/REVISION`, which does have a wrong value.

This file is not stored in the repository and is created automatically by `contrib/gcc_update`.

Later, the Advance Toolchain build manipulates this file on GCC stage 1.  The changes from this step are not working anymore.

So, we likely have 2 issues there:
1. `contrib/gcc_update --touch` needs to be executed after the checkout to guarantee we have the correct revision on 'gcc/REVISION'.
2. It's necessary to fix the pre hacks of GCC stage 1.

er-1:
gcc/REVISION may not match with the requested revision.
In revision 259582, gcc/REVISION contains 259562.

contrib/gcc_update --touch doesn't update gcc/REVISION.
contrib/gcc_update updates gcc/REVISION but it pulls newer files which means it's not the wanted revision any longer.

$ gcc-7-branch> svn info
Path: .
Working Copy Root Path: /home/erwan/tmp/at-build-tray/sources/gcc-7-branch
URL: svn://gcc.gnu.org/svn/gcc/branches/ibm/gcc-7-branch
Relative URL: ^/branches/ibm/gcc-7-branch
Repository Root: svn://gcc.gnu.org/svn/gcc
Repository UUID: 138bc75d-0d04-0410-961f-82ee72b054a4
Revision: 259582
Node Kind: directory
Schedule: normal
Last Changed Author: pc
Last Changed Rev: 259582
Last Changed Date: 2018-04-23 18:14:38 -0300 (Mon, 23 Apr 2018)

$ gcc-7-branch> cat gcc/REVISION 
[ibm/gcc-7-branch revision 259562]

I propose this: after the checkout, take the revision number from the svn output to update gcc/REVISION.

tuliom:
The GCC team is clarifying why gcc/REVISION was checked in and what that revision number means.
I like your proposal, but let's wait for their reply.  I suspect the revision number in that file will help to fix the issue 2, where both the revision of branches/ibm/gcc-7-branch and branches/gcc-7-branch are listed.

er-1:
[...]

PR #498, opened 2018-06-21
Update the display of gcc -v

er-1:
`gcc -v` will print the FSF revision number the IBM branch in based upon, and the IBM branch revision number.


configs/11.0/packages/gcc/stage_1
```diff
@@ -31,8 +31,8 @@ atcfg_pre_hacks() {
 	echo "Advance-Toolchain-${at_major_internal}" > ${ATSRC_PACKAGE_WORK}/gcc/DEV-PHASE
 	# Overwrite the existing REVISION with the current SVN revision date
 	mv ${ATSRC_PACKAGE_WORK}/gcc/REVISION ${ATSRC_PACKAGE_WORK}/gcc/REVISION.back
-	sed "s/branch merged/branch, revision: ${ATSRC_PACKAGE_REV} merged/" ${ATSRC_PACKAGE_WORK}/gcc/REVISION.back \
-		 > ${ATSRC_PACKAGE_WORK}/gcc/REVISION
+	sed "s/ \([^ ][^ ]*\)]$/ ${ATSRC_PACKAGE_REV}, based upon FSF revision \1]/" \
+	        ${ATSRC_PACKAGE_WORK}/gcc/REVISION.back > ${ATSRC_PACKAGE_WORK}/gcc/REVISION
```
tuliom: I think this code needs a mechanism to detect when sed does not match anything.  This way we would catch this kind if issues earlier.
Maybe a diff between REVISION.back and REVISION?

tuliom: Does AT next also come with a gcc/REVISION or do we need to generate it?

The older AT versions (at least 9.0 and 10.0) also need this fix.

er-1:
@tuliom as asked I've restored the symlink between next and AT 11.0.

er-1:
Fixes #497

Issue #511, opened 2018-06-26
Valgrind itrace from AT 8.0 not building on newer kernels


tuliom:
Valgrind from AT 8.0 does not build on Linux 4.x.
The following change should solve the issue:

```
diff -paurN valgrind-3.10.0/configure.ac vg-310-itrace/configure.ac
--- valgrind-3.10.0/configure.ac	2014-09-11 04:12:44.000000000
-0500
+++ vg-310-itrace/configure.ac	2014-11-12 10:43:27.598185054
-0600
@@ -303,7 +303,7 
         kernel=`uname -r`
 
         case "${kernel}" in
-             2.6.*|3.*|4.*) 
+             2.6.*|3.*) 
         	    AC_MSG_RESULT([2.6.x/3.x family (${kernel})])
         	    AC_DEFINE([KERNEL_2_6], 1, [Define to 1 if you're
using Linux 2.6.x or Linux 3.x])
         	    ;;
```

PR #512, opened 2018-06-29
Take and apply a patch for Valgrind (AT 8.0)

er-1:
The patch allow Valgrind 3.10 to be built with kernel 4.*.


er-1:
Fixes #511

PR #527, opened 2018-07-07
Update glibc on AT next

Cherbot:
Bump to revision 916124ed8417


tuliom:
I suggest to drop Ubuntu 14.04 from the distros testing AT next.  We don't support it anyway.

PR #528, opened 2018-07-07
Update expat on AT next

Cherbot:
Bump to revision 491aa5cbf515


er-1:
From configure of glibc, make is too old:
```
configure:4703: checking version of make
configure:4713: result: 3.81, bad
[...]
configure:5289: error: 
*** These critical programs are missing or too old: make
*** Check the INSTALL file for required versions.
```

tuliom:
The expat update should not be in the same branch of the glibc one.  There is something wrong here.

PR #530, opened 2018-07-07
Update binutils on AT next

Cherbot:
Bump to revision 674e55f7b11e


configs/next/packages/binutils/sources
```diff
@@ -21,7 +21,7 @@
 
 ATSRC_PACKAGE_NAME="GNU Binutils"
 ATSRC_PACKAGE_VER=2.30.51
```
tuliom: The version needs to be changed to 2.31.51.

er-1:
PR #539 overrides this one.

PR #531, opened 2018-07-07
Update openssl on AT next

Cherbot:
Bump to revision c9d6fdd6f79c


configs/next/packages/openssl/sources
```diff
@@ -21,7 +21,7 @@
 
 ATSRC_PACKAGE_NAME="Open Secure Sockets Layer"
 ATSRC_PACKAGE_VER="1.1.1-pre7"
```
tuliom: The version should be 1.1.1-pre8.

PR #537, opened 2018-07-10
Diverge AT 12.0 from next

tuliom:
This patchset has the initial steps to diverge AT 12.0 from next.

It has already diverged the following packages:
 - gcc
 - glibc
 - binutils
 - OpenSSL
 - GDB

All the other packages still have to be reviewed and updated accordingly.


configs/next/build.mk
```diff
@@ -0,0 +1 @@
+../11.0/build.mk
```
gftg85: Why not point to ../12.0/build.mk?

gftg85:
OK with the changes.

configs/next/build.mk
tuliom: It should.... It seems `make clone` has a bug.

tuliom:
> Did you mean 12.0?

@gftg85 No.  AT next does indeed have version 13.0.  Don't you agree?

tuliom:
@gftg85 Fixed the symlink pointing to AT 11.0.

gftg85:
>AT next does indeed have version 13.0. Don't you agree?
I see what you mean, now.  Maybe the use of the word "Point" in your sentence is a bit misleading (since it could give the impression that it refers to symlinks, which point to 12.0), but not a big deal and fine the way it is.

PR #538, opened 2018-07-12
Update openssl on AT next

er-1:
Bump to version 1.1.1-pre8


er-1:
@tuliom thanks for the advice. I updated `ATSRC_PACKAGE_REV` as well.

er-1:
@tuliom I squashed my commits into one.

PR #552, opened 2018-07-18
Update more packages on AT 12.0

gftg85:


configs/12.0/packages/oprofile/sources
```diff
@@ -20,7 +20,7 @@
 #
 ATSRC_PACKAGE_NAME="OProfile"
 ATSRC_PACKAGE_VER=1.3.0
-ATSRC_PACKAGE_REV=21f415c04019
+ATSRC_PACKAGE_REV=76c4275a0cb5
```
tuliom: AFAICS this commit ID is newer than 21f415c04019.

Why do you need to leave AT next in an older checkout?
Can't it continue as a symlink to AT 12.0?

I should have put this comment in the AT next file, but Github doesn't allow me.

configs/12.0/packages/libhugetlbfs/sources
```diff
@@ -19,8 +19,8 @@
 # ==========================================
 #
 ATSRC_PACKAGE_NAME="Libhugetlbfs"
-ATSRC_PACKAGE_VER=2.21
-ATSRC_PACKAGE_REV=02df38e93e25
+ATSRC_PACKAGE_VER=2.20
+ATSRC_PACKAGE_REV=e44180072b79
```
tuliom: If libhugetlbfs is downgraded, doesn't it require to diverge AT next from AT 12.0?

configs/12.0/packages/oprofile/sources
gftg85: Honestly, I don't think that _none_ of the `sources` files on AT Next should be symlinks to stable versions of AT.  AT Next gets automatically updated every week, and I wouldn't like to have to check if the automatic updates to AT Next are affecting AT 12.0 (or any other stable version, for that matter).

While it is true that oprofile on AT Next could be updated, I think it's out of the scope for this pull request.

tuliom: You're completely right.  I didn't take into account the automatic updates.

configs/12.0/packages/libhugetlbfs/sources
gftg85: On master, `libhugetlbfs/sources` is already a regular file (it has been converted by commit ID db8fdbc565bb)

PR #555, opened 2018-07-21
Update gcc on AT next

Cherbot:
Bump to revision 8f2f3eeb9e4c


er-1:
In glibc_1-64 build logs:
`../o-iterator.mk:9: recipe for target 'glibc_1-64/posix/fnmatch.os' failed`

tuliom:
Which is [this bug](https://sourceware.org/bugzilla/show_bug.cgi?id=23442), fixed in glibc commit 9c79cec8cd2a.
We have to wait until the next glibc update, rebase this patch and confirm if it works.

PR #557, opened 2018-07-21
Update mpfr on AT next

Cherbot:
Bump to revision 12945


er-1:
Same issue as in PR #553.

PR #563, opened 2018-07-23
Update Python on AT >= 11.0

tuliom:
Update Python to version 3.6.6.


tuliom:
AT next Ubuntu16 ppc64le 868 failed while creating the deb package.
I don't think the issue is related to this patch, so I requested a new build.

tuliom:
I requested a new build for AT 9.0 RHEL7 ppc64le.

PR #564, opened 2018-07-23
[RFC] Downgrade OpenSSL to a stable version on AT next

tuliom:
OpenSSL is a dependency for other packages, specially Python.
The relationship between both packages has been known to cause an extra
burden, i.e. we have a list of Python tests that are ignored
specifically because they're known to fail with certain OpenSSL revisions.
The extra burden increased when AT next started to track OpenSSL's
master branch and its frequent changes.  These frequent modifications
may cause issues that are not instantly fixed for obvious reasons.

This patch is intended to stop the Whac-a-mole behavior and downgrade
OpenSSL to a stable branch on AT next.


gftg85:
I'm generally against `sources` files on AT Next being symlinks to `sources` files on __any__ stable versions of AT, because of the automatic updates to AT Next.  Such automatic updates would be proposed, by Cherbot, as pull requests with a title similar to "Update OpenSSL on AT Next", which would be true, but a bit misleading.

Unless you're planning to remove the automatic updates of OpenSSL on AT Next.

PS: I tried to add this message as an inline comment, but I can't because Github won't let me make inline comments to symlinks.

tuliom:
@gftg85 I do plan to disable the automatic update of OpenSSL for AT next on Cherbot.

tuliom:
Everyone agrees with this RFC.
So, I'm merging this patch as it fixes a regression with commit a5d8a7dbcdfd0c710358921a7e91b86cb6916763.

PR #568, opened 2018-07-25
Add missing RPM requirements

### LINKED ISSUE #194: Missing required packages
The installation process run some programs which aren't listed as requirements.
They're usually installed in most of the minimum distro installs, but they aren't available in contained environments, e.g. mock.

It's necessary to list at least gawk, grep, sed and systemd.

er-1:
Required a few packages that are run during installation.
Those packages are part most of the time of minimum distro install
but they could be missing.

Fixes #194 


er-1:
@tuliom I updated the patch.

tuliom:
The code looks good to me, but let's wait for the results of the tests before merging it.  ;-)

er-1:
Fixes #194

PR #569, opened 2018-07-26
Add requirement for GNU make on AT 12.0

er-1:
GNU make 4.0 (or greater) is required by glibc 2.28 which is part
of AT 12.0 (or greater).


er-1:
The build on Ubuntu16 failed because the function, `check_packages`,  that checks the required packages (from `AT_COMMON_PKGS_REQ` for example) doesn't get the same kind of list from RedHat/SuSE systems than Ubuntu/Debian systems.
On RedHat/SuSE systems, the list is made of `<package name>-<version>`.
On Ubuntu/Debian systems, the list is made of `<package name>`.
This means if `AT_COMMON_PKGS_REQ` has a version requirement for a package, this version requirement will fail when comparing with the list of `<package name>` only.

er-1:
This PR needs PR #571 to go further.

configs/12.0/distros/ubuntu-14.mk
```diff
@@ -70,7 +70,8 @@ ifndef AT_DISTRO_REQ_PKGS
     AT_COMMON_PKGS_REQ := zlib1g-dev libncurses5-dev ncurses-term flex bison \
                           texinfo subversion cvs gawk fakeroot debhelper \
                           autoconf rsync curl bc libxml2-utils automake \
-                          docbook2x dpkg-sig xutils-dev wget autoconf-archive
+                          docbook2x dpkg-sig xutils-dev wget autoconf-archive \
+                          make-[4-9]
```
RajalakshmiS: @er-1  If make-[4-9]  means make-4 to make-9 .. why should we restrict till make-9 ? I think our intention is to check make 4+ .

tuliom: I do agree this is the correct solution, but I don't think we need to be over-engineer this.
I suggest the following:
1. Remove ubuntu-14.mk.  We don't support it on AT >= 12.0.  We don't even test it.
2. For SLES 12 and Ubuntu 16.04, we can have just `make` because these distros already provide a GNU make that's newer than 4.0.
3. If you really believe that we need to specify a version there, `[4-9]` should be a large enough window, i.e. Ubuntu 16.04 will likely reach EOL before GNU make 10.0 is released.  ;-)

RajalakshmiS: @tuliom Ack

er-1: @tuliom I'll remove ubuntu-14.mk. What about centos-6.mk and centos-7.mk? I guess I could remove them too.

er-1:
I've updated the patch.

PR #571, opened 2018-07-26
Fix the list of packages to compare requirements with

### LINKED ISSUE #570: check_packages fails on Ubuntu/Debian
The function `check_packages` that checks the required packages (from `AT_COMMON_PKGS_REQ` for example) doesn't get the same kind of list from RedHat/SuSE systems than Ubuntu/Debian systems.
On RedHat/SuSE systems, the list is made of `<package name>-<version>`.
On Ubuntu/Debian systems, the list is made of `<package name>`.
This means if `AT_COMMON_PKGS_REQ` has a version requirement for a package, this version requirement will fail when comparing with the list of `<package name>` only.

er-1:
On RedHat/SuSE, the list is made of package name+version.
However, on Ubuntu/Debian the list was made only of package name.
This meant a version requirement, like make-[4-9], made the check
failed while comparing with the list of package name only.

Fixes #570 

### DIFF
--- a/scripts/helpers/utilities.mk
+++ b/scripts/helpers/utilities.mk
@@ -671,7 +671,7 @@ define check_packages
         missing_pkg="no"; \
         tmp_file=$$(mktemp); \
         if [[ ("$(DISTRO_FM)" == "ubuntu") || ("$(DISTRO_FM)" == "debian") ]]; then \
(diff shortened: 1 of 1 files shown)


er-1:
Closes #570

Issue #573, opened 2018-07-30
FVTR: ck_ldds should not check the interpreter of debug info files


tuliom:
Debug info files are not complete executables and do not have the exact same information available in shared libraries or executables.
So, ck_ldds should not check these files because we usually see this:
```
[Requesting program interpreter: ]
```

tuliom:
Fixed.

Issue #143, opened 2017-08-09
[FVTR] gccgo test is failing on pcc64 for AT >= 10.0


rzinsly:
The gccgo test is failing on AT 11.0 when built for ppc64 (BE):
gccgo: power8:  Running gccgo package test...
t^M
Hello!^M
fatal error: unexpected signal during runtime execution^M
gccgo: power8:  gccgo test failed

By manually running we got the following errors:
./gopkgtest                                                                                              
Hello!
fatal error: unexpected signal during runtime execution
[signal SIGSEGV: segmentation violation code=1 addr=1731642852816977920 pc=70367025496785]

goroutine 16 [running]:
runtime.dopanic
        /home/rzinsly/AT/tmp/at11.0-0-beta1.redhat-7_ppc64_ppc64/sources/gcc/libgo/go/runtime/panic.go:832
runtime.throw
        /home/rzinsly/AT/tmp/at11.0-0-beta1.redhat-7_ppc64_ppc64/sources/gcc/libgo/go/runtime/panic.go:758
runtime.sigpanic
        /home/rzinsly/AT/tmp/at11.0-0-beta1.redhat-7_ppc64_ppc64/sources/gcc/libgo/go/runtime/signal_unix.go:241
runtime.sighandler
        /home/rzinsly/AT/tmp/at11.0-0-beta1.redhat-7_ppc64_ppc64/sources/gcc/libgo/go/runtime/signal_sighandler.go:59
runtime.sigtrampgo
        /home/rzinsly/AT/tmp/at11.0-0-beta1.redhat-7_ppc64_ppc64/sources/gcc/libgo/go/runtime/signal_unix.go:227

        :0

        :0

        :0

        :0
time.formatNano
        /home/rzinsly/AT/tmp/at11.0-0-beta1.redhat-7_ppc64_ppc64/sources/gcc/libgo/go/time/format.go:403
time.AppendFormat.N9_time.Time
        /home/rzinsly/AT/tmp/at11.0-0-beta1.redhat-7_ppc64_ppc64/sources/gcc/libgo/go/time/format.go:612
time.Format.N9_time.Time
        /home/rzinsly/AT/tmp/at11.0-0-beta1.redhat-7_ppc64_ppc64/sources/gcc/libgo/go/time/format.go:457
time.String.N9_time.Time
[...]

PR #574, opened 2018-07-31
Do not check the interpreter of debug files in ck_ldds

tuliom:
Debug files may not have information about the interpreter and can be
safely ignored when validating the interpreter.

Fixes issue #573.


tuliom:
Build AT 10.0 RHEL7 ppc64le 9 failed because of issue #143 .
Build AT 9.0 SLES12 ppc64le 380 is also unrelated to this patch.  It failed after a server issue.

PR #584, opened 2018-08-04
Update valgrind on AT next

Cherbot:
Bump to revision b9cfb2d15413


configs/next/packages/valgrind/sources
```diff
@@ -21,7 +21,7 @@
 
 ATSRC_PACKAGE_NAME="Valgrind"
 ATSRC_PACKAGE_VER=3.14.0
```
gftg85: On the one hand, this is AT Next, so I don't see a problem with stating that this is Valgrind 3.14 (which has never been released).  On the other hand, a question came to my mind:

Why do we set version information of the packages in AT Next.  This means extra work for us (because we have to check if there was a version bump), with no benefits (at least, I don't see any).  So, should we drop version information from the packages in AT Next?

PR #588, opened 2018-08-04
Update expat on AT 12.0

Cherbot:
Bump to revision e113a4667258


tuliom:
Cherbot has been modified. Expat won't be updated automatically on AT 12.0 anymore.

PR #604, opened 2018-08-18
Update glibc on AT next

Cherbot:
Bump to revision 86a6c75a2948


configs/next/packages/glibc/sources
```diff
@@ -21,7 +21,7 @@
 
 ATSRC_PACKAGE_NAME="GNU C Library"
 ATSRC_PACKAGE_VER=2.28.9000
-ATSRC_PACKAGE_REV=e5721f45f637
+ATSRC_PACKAGE_REV=86a6c75a2948
```
gftg85: OK.

$ git show 86a6c75a2948:version.h | grep VERSION
#define VERSION "2.28.9000"

PR #616, opened 2018-08-22
Serialize locale generation for AT >= 9.0

tuliom:
Avoid random build failures by serializing locale generation.
This patch may increase build time by 5 minutes.


tuliom:
I created this patch a while ago when I noticed that locale generation was failing randomly in some servers.
However, this is not happening anymore and I believe this patch should not be applied (yet).
I'm closing this PR.

gftg85:
I mentioned (in pull request #621) that this change might actually be useful, and @tuliom apparently aggrees, so I'll reopen and review.

PR #619, opened 2018-08-25
Update liburcu on AT next

Cherbot:
Bump to revision 34b11dd41c61


configs/next/packages/liburcu/sources
```diff
@@ -21,7 +21,7 @@
 
 ATSRC_PACKAGE_NAME="Userspace RCU"
 ATSRC_PACKAGE_VER=0.10.1
-ATSRC_PACKAGE_REV=ceb592f99ca8
+ATSRC_PACKAGE_REV=34b11dd41c61
```
gftg85: This update is not exactly OK...

Version 0.10.1 is in a stable branch (stable-0.1), but AT Next follows the master branch.

The next version will probably be tagged as 0.11.0, as can be seen with:

```
$ git show 34b11dd41c61:configure.ac | grep AC_INIT
AC_INIT([userspace-rcu],[0.11.0-pre] (...)
```

This was already wrong even befor this update, as can be seen with:

```
$ git show ceb592f99ca8:configure.ac | grep AC_INIT
AC_INIT([userspace-rcu],[0.11.0-pre] (...)
```

In my opinion, it it OK to merge this pull request, but we should change ATSRC_PACKAGE_VER to something else.
Maybe we should start a discussion about this field on AT Next, as suggested in #584.

tuliom: @gftg85 I started this discussion in PR #640.

Issue #8, opened 2017-02-16
The loader of the cross compiler is referencing the wrong ld.so.cache


rzinsly:
When running the cross compiler's /opt/atX.X/ppc*/sbin/ldconfig from QEMU user mode, it tries to update /etc/ld.so.cache instead of /opt/atX.X/ppc*/etc/ld.so.cache.

Note: **be careful with these tests, because they may break your system!** It's better to run them on a guest VM. Remember to create a snapshot of the VM before running them.

PR #624, opened 2018-08-29
Fix cross compiler's ld.so.cache path

YasminBeatriz:
Set the path of ld.so.cache to `/opt/atX.X/ppc*/etc` via glibc's `--sysconfdir` build option. The file must exist, so we're also reverting its removal in the post install stage.


tuliom:
This PR is meant to fix issue #8.

pjreid819:
Any specific reason this has not been merged in yet?

PR #630, opened 2018-09-08
Update glibc on AT next

Cherbot:
Bump to revision 38245425a9ad


tuliom:
@er-1 I'm not sure I agree.  The initial problem was related to using the system's C++ compiler.
The commit you mentioned is changing how tests are looking for the loader.
I've been monitoring this discussion, but I haven't seen progress recently.

er-1:
@tuliom We're right, it doesn't fix the issue. I gave a try, it failed.

PR #651, opened 2018-09-29
Update binutils on AT next

Cherbot:
Bump to revision 6424bdfcfc3a


tuliom:
Python's test_signal failed on builds RHEL7 ppc64le 112 and SLES12 ppc64le 943.
I don't think these failures are related to this update, so I restarted the builds.

PR #652, opened 2018-09-29
Update gcc on AT next

Cherbot:
Bump to revision fe0942b8c864


tuliom:
Python's test_signal failed on build RHEL7 ppc64le 111.
I don't think this failure is related to this update.  I restarted the build to confirm.

PR #654, opened 2018-09-29
Update mpfr on AT next

Cherbot:
Bump to revision 13233


er-1:
> This revision is broken

It seems it has been fixed:
```
r13235 | vlefevre | 2018-10-01 10:51:10 -0300 (Mon, 01 Oct 2018) | 8 lines
Changed paths:
   M /trunk/configure.ac
   M /trunk/src/Makefile.am

[configure.ac,src/Makefile.am] Avoid a "make" failure due to a bug in
Automake 1.15 automatic dependency tracking: instead of using mini-gmp
directly in an external directory with a $(mini_gmp_path) variable in
src/Makefile.am, make the configure script add symbolic links to the
mini-gmp sources in the build directory, and use these symbolic links
for the MPFR build (as if they were plain files).
Tested under various conditions: static and shared; separate source and
build directories; changing the mini-gmp version without "make clean".
```

tuliom:
@er-1 Nice catch!  Let's close this PR and wait for the next one.

PR #668, opened 2018-10-06
Update openssl on AT 9.0

Cherbot:
Bump to revision fff1da43be22


tuliom:
Build AT 9.0 SLES12 ppc64le 398 failed because of python's test_epoll.
I don't think the issue is related.  I'm restarting the build.

gftg85:
>Build AT 9.0 SLES12 ppc64le 398 failed because of python's test_epoll.
I don't think the issue is related. I'm restarting the build.

You were right, thanks.

PR #695, opened 2018-10-20
Newer openssl and python versions on AT Next

gftg85:
While trying to update OpenSSL to the latest stable branch (the branch for version 1.1.1) on AT Next, I got a single FVTR failure, precisely in Python's test_ssl.

In order to check if newer Python versions had already been patched to work correctly with newer OpenSSL, I also updated Python to its latest stable branch (the branch for version 3.7).  This required a couple of downstream patches, but worked correctly in a SLES 12 test that I started manually.

This pull request combines all these changes and I hope that it passes the continous integration tests.  Fingers crossed.

**Note**: I recommend reviewing each patch in this set individually, because the first of them only changes file names.  Looking at the combined diff is useless.


gftg85:
Fixed typo in the requirements for RHEL 7.  `autoconf-archivep` -> `autoconf-archive`

tuliom:
@gftg85 said in the commit message:
>     This patch also adds a patch to disable the checking of the version of
>     zlib in test_zlib.  This is required because the version of zlib that AT
>     Next distributes contains a non-numeric suffix.

Does it affect stable AT releases too?
Because if they don't need this patch, we may want to remove it before creating the AT 13.0 config set (or 14.0, etc.).  In other words, we may want a comment in configs/next/packages/python/sources:66 explaining this patch is only required in AT next and could be safely removed from stable versions.

gftg85:
>Does it affect stable AT releases too?
Because if they don't need this patch, we may want to remove it before creating the AT 13.0 config set (or 14.0, etc.). In other words, we may want a comment in configs/next/packages/python/sources:66 explaining this patch is only required in AT next and could be safely removed from stable versions.

It affects stable releases, too.

**Current** stable releases do not **currently[1]** have this problem, because they fetch Python from a commit ID that does not contain the failing check for zlib version.  We can verify that with the following command (the hash is the revision that AT 11.0 and 12.0 get from Python):

```
$ git show 4cf1f54eb764:Lib/test/test_zlib.py | grep "tuple.*ZLIB_RUNTIME_VERSION"

(no output)
```

On AT Next, which gets a newer revision from Python sources, the output looks like this:

```
$ git show 657e3f9a2c0d:Lib/test/test_zlib.py | grep "tuple.*ZLIB_RUNTIME_VERSION"
        ver = tuple(int(v) for v in zlib.ZLIB_RUNTIME_VERSION.split('.'))
```

This means that AT 13.0, when it becomes stable, will need this downstream patch.


[1] Although I think I have properly answered your question, I hadn't previously noticed that the stable branch for Python 3.6 also has the check for zlib version (added by commit ID 73653062b1ef).  We don't currently see this problem on AT 11.0 and 12.0, because we use an older commit.  When Python 3.6.7 gets released, and AT11.0 and 12.0 get updated, we might need this downstream patch for them.  So, thanks for bringing this up.  Anyhow, I don't believe that this assessment should block this pull request.

gftg85:
>@gftg85 have you already created patches for cherbot to automatically update OpenSSL and Python on AT next?

I haven't.

gftg85:
>This patch also adds a patch to disable the checking of the version of
zlib in test_zlib.  This is required because the version of zlib that AT
Next distributes contains a non-numeric suffix.

My [comment](https://github.com/advancetoolchain/advance-toolchain/pull/695#issuecomment-431999296) explained that the problem happens on AT Next because the test case in Python changed (not because of any changes in zlib) so this comment looks wrong, now.

If I change this comment to the following paragraph, does it look better and OK for master?

```
This patch also adds a patch to disable the checking of the version of
zlib in test_zlib.  This is required because, in the newer version of 
Python used by AT Next, the checking of zlib version does not accept  
non-numeric components, but zlib on AT 12.0 and Next contains a       
non-numeric suffix ('-motley').
```

tuliom:
I don't think it's necessary. It was just a misunderstanding from my part. LGTM

PR #697, opened 2018-10-27
Update gcc on AT next

Cherbot:
Bump to revision 301a5923ff56


tuliom:
I'm seeing reports of this warning appearing when using GCC 8.2. It looks like it was backported to GCC 8.2 and will soon affect AT 12.0.
I suspect we'll need a patch to build Binutils with `-Wno-error=class-memaccess`.

tuliom:
This issue is blocking a GCC update, which blocks glibc updates (PR #714).

er-1:
This is overriden by PR #746.

Issue #649, opened 2018-09-26
HTML files in docs/ do not pass a validation test


tuliom:
All the files in `docs/` are failing HTML validation, e.g.: `tidy -config docs/tidy.cfg docs/advance-toolchain.html`

It's necessary to fix all of them:
- [x] docs/advance-toolchain.html
- [ ] docs/advtool-compiler.html
- [ ] docs/advtool-faq.html
- [ ] docs/advtool-installation.html
- [ ] docs/adv-tool-usage.html

tuliom:
`docs/tidy.cfg` is missing from the repository.
@tuliom should commit it.

PR #713, opened 2018-11-08
Fix HTML syntax and tags

er-1:
Fix HTML syntax and tags to get rid of errors reported by `tidy`.

### DIFF
--- a/docs/adv-tool-usage.html
+++ b/docs/adv-tool-usage.html
@@ -26,7 +26,7 @@
 
 <!-- Start the get support button and overlay -->
 <div class="center">
-   <a href="#" onclick="IBMCore.common.widget.overlay.show('SUPPORT'); return false;">Get support</a></p>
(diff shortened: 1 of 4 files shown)


gftg85:
@er-1,

In #649, there's a mention do a configuration file (`docs/tidy.cfg`), but the file is not in the repository.  Do you have access to it?  Without it, I can't run `tidy` as reported in the bug, thus I can't verify that it actually fixes #649.

er-1:
@gftg85,
Good catch.
My fix is based on running `tidy` without this configuration file `docs/tidy.cfg`. I agree it might not report the same errors/warnings.

gftg85:
OK.

Your fixes look good to me, but the statement `Fix #649`, in the pull request message, would automatically close the bug.

Could you commit it directly (i.e.: not using the github.com interface), please?

er-1:
@gftg85 
I removed the statement `Fix #649`, as we can't ensure this PR fixes the related issue (due to missing `docs/tidy.cfg`).

gftg85:
Good, now you can safely merge using the github.com interface.  Thanks.

PR #714, opened 2018-11-17
Update glibc on AT next

Cherbot:
Bump to revision c75772e3f079


gftg85:
Failed due to errors similar to these:

```
errno-loc.c:28:1: error: ‘copy’ attribute directive ignored [-Werror=attributes]
28 | libc_hidden_def (__errno_location)
   | ^~~~~~~~~~~~~~~
version.c:54:1: error: ‘copy’ attribute directive ignored [-Werror=attributes]
54 | weak_alias (__gnu_get_libc_release, gnu_get_libc_release)
   | ^~~~~~~~~~
version.c:62:1: error: ‘copy’ attribute directive ignored [-Werror=attributes]
62 | weak_alias (__gnu_get_libc_version, gnu_get_libc_version)
   | ^~~~~~~~~~
```

which are probably due to the following commit

```
commit 1626a1cfcd6ba1cc64721be1036c2873211d499b
Author: Martin Sebor <msebor@redhat.com>
Date:   Fri Nov 9 17:24:12 2018 -0700

    Add support for GCC 9 attribute copy.
```

Several other occurrences of similar errors have already been fixed by Joseph Myers, such as the fix in the following commit

```
commit 9c4b457e1b47369a40afd5e5e220f5d657667c0a
Author: Joseph Myers <joseph@codesourcery.com>
Date:   Tue Nov 13 01:53:53 2018 +0000

    Add hidden_tls_def macros, fix powerpc-soft build with GCC 9.
```

We might need a similar fix for the advance-toolchain.

tuliom:
> We might need a similar fix for the advance-toolchain.

@gftg85 Do you mean upstream glibc instead of advance-toolchain?

If I'm understanding this right, glibc can't be built for ppc64/ppc64le using GCC 9.
In other words, AT next cannot update glibc until we fix this.

gftg85:
>>We might need a similar fix for the advance-toolchain.
>
>Do you mean upstream glibc instead of advance-toolchain?

I didn't mean that when I wrote my comment.  Anyhow, you could be right to say that this might need a patch in upstream glibc, but I'm not sure.

>If I'm understanding this right, glibc can't be built for ppc64/ppc64le using GCC 9.

That doesn't seem to be totally true.  I can build current glibc master with current gcc trunk.  This error seems like a problem that the advance-toolchain is exposing (maybe a newer gcc **on the advance-toolchain (next)** could fix the problem?).  We need to investigate.

>In other words, AT next cannot update glibc until we fix this.

Yes, but not necessarily in upstream glibc or gcc, it could be a problem on the advance-toolchain.

tuliom:
> This error seems like a problem that the advance-toolchain is exposing (maybe a newer gcc on the advance-toolchain (next) could fix the problem?).

OK.  I get your point now.

AT next distributing GCC 9, but an older revision that doesn't provide the copy attribute, breaking the code that checks for `#if __GNUC_PREREQ (9, 0)`.

In other words: we can only update glibc after PR #697 is fixed.

PR #718, opened 2018-11-17
Update tbb on AT next

Cherbot:
Bump to revision 8ff3697f544c


configs/next/packages/tbb/sources
```diff
@@ -20,8 +20,8 @@
 #
 
 ATSRC_PACKAGE_NAME="Thread Building Blocks"
-ATSRC_PACKAGE_VER=2019_U1
-ATSRC_PACKAGE_REV=4cebdd9a1c9f
+ATSRC_PACKAGE_VER=2019_U2
+ATSRC_PACKAGE_REV=8ff3697f544c
```
gftg85: Now fixed.

tuliom: @gftg85 The web interface created another commit.  Is this what you expected?

The Ubuntu 16.04 test failed with:
```
dh_install --fail-missing
libfakeroot, when sending message: Invalid argument
debian/rules:33: recipe for target 'binary-arch' failed
make[1]: *** [binary-arch] Error 1
```

I restarted the test.

gftg85: >The web interface created another commit. Is this what you expected?

@tuliom, I actually like the fact that it creates a new commit, because it shows that Cherbot can only go so far...  On the other hand, it would make it hard to bisect if the first commit would cause builds to fail (as is the case with gcc version bumps). So, maybe they should be squashed? (I don't have a strong opinion about it, nor about the gcc case).

I see that the option to squash commits using the web interface is disabled (which is a good thing, in my opinion), so this should probably be squashed and pushed from the command-line, once the pull request is approved, and if we decide that we want to squash the commits.

tuliom: > I don't have a strong opinion about it, nor about the gcc case

I do have have a strong opinion against commits that break bisect (i.e. gcc case), but I do not oppose to the TBB case.  I was just confirming it was intentional.  ;-)

gftg85: @tuliom, Just to be clear, are you OK with this being merged as-is?

tuliom: @gftg85 yes.

PR #733, opened 2018-11-24
Update liburcu on AT 12.0

Cherbot:
Bump to revision 6c249cd8c0a4


er-1:
@tuliom 
I'll do that.
I've noticed `stage_1` could be also replaced with a symlink.
What do you think?

PR #747, opened 2018-12-04
Update link to gcc documentation on AT 12.0

### LINKED ISSUE #736: The release notes from AT 12.0-0 points to the wrong GCC documentation
The correct link is https://gcc.gnu.org/onlinedocs/gcc-8.2.0/gcc/

The community does not maintain a page for GCC 8.2.1.
We have to replace the last digit with 0.

er-1:
There is no documentation for the current version of gcc8, 8.2.1.
The link has been updated to refer to the version 8.2.0 which has documention.

Closes #736 


configs/12.0/packages/gcc/sources
```diff
@@ -25,7 +25,7 @@ ATSRC_PACKAGE_SUBPACKAGE_NAME_1="GNU Libgomp"
 ATSRC_PACKAGE_VER=8.2.1
 ATSRC_PACKAGE_REV=266159
 ATSRC_PACKAGE_LICENSE="GPL 3.0"
-ATSRC_PACKAGE_DOCLINK="https://gcc.gnu.org/onlinedocs/gcc-${ATSRC_PACKAGE_VER}/gcc/"
+ATSRC_PACKAGE_DOCLINK="https://gcc.gnu.org/onlinedocs/gcc-8.2.0/gcc/"
```
tuliom: What about `${ATSRC_PACKAGE_VER/%.[0-9]/.0}` instead?
We should also add a comment here explaining why this is necessary.

er-1: @tuliom Good catch. I've updated my patch and added a comment.

PR #756, opened 2018-12-11
Modify the protocol to clone GCC git tree

er-1:
Using https protocol instead of git protocol as a workaround to get
rid of a network limitation that prevents the git tree to be cloned.


er-1:
This is facing errors similar to those in PR #714

tuliom:
We need to investigate this closely.
I didn't expect the same kind of errors of PR #714 without updating glibc first.

tuliom:
I've just confirmed the build passes if we update glibc, gcc and Binutils together.
I'm preparing a patch.

PR #773, opened 2019-01-05
Update mpfr on AT 12.0

Cherbot:
Bump to revision 13363


er-1:
The builds failed because of a glitch with SVN (local repository locked).

PR #779, opened 2019-01-15
Update Binutils, GCC and glibc on AT next

tuliom:
In the past 2 months, we have identified a series of package updates in Binutils, GCC and glibc that are interdependent.
This pull request is meant to update them and close all the previous pull requests that were blocked.
Closes #714 , #756 , #758 , #774 .

PR #792, opened 2019-01-22
Update Python on AT next

tuliom:
Bump revision to 9a3ffc0492d1.
Update the getlib64 patch in order to continue working.

This patch fixes the issue with Python's test_email found in PR #779 .

### DIFF
--- a/configs/next/packages/python/sources
+++ b/configs/next/packages/python/sources
@@ -21,7 +21,7 @@
 
 ATSRC_PACKAGE_NAME="Python"
 ATSRC_PACKAGE_VER=3.7
-ATSRC_PACKAGE_REV=657e3f9a2c0d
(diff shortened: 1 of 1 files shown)


er-1:
It failed with applying a patch on gcc.
With gcc 9.*, it needs the following patch: https://raw.githubusercontent.com/powertechpreview/powertechpreview/e49f20bdf5e310b3d962ee683c3e05e88d2a0225/GCC%20PowerPC%20Backport/6/0001-libssp-Ignore-vsnprintf-test-when-ssp_have_usable_vs.patch
It's part of PR #756.
Is there a way to include this PR to PR #779 ? As PR #756.

tuliom:
> Is there a way to include this PR to PR #779 ? As PR #756.

Yep. It seems I won't be able to split this while AT next is failing.
Changes to the other PR will likely fail too.

PR #807, opened 2019-02-07
Enable the python github repository on AT next

er-1:
On AT next, enable the python github repository. This allows Cherbot to
check, on a weekly basis, if there are available updates.


configs/next/packages/python/sources
```diff
@@ -21,16 +21,16 @@
 
 ATSRC_PACKAGE_NAME="Python"
 ATSRC_PACKAGE_VER=3.7
-ATSRC_PACKAGE_REV=9a3ffc0492d1
+ATSRC_PACKAGE_REV=27f6e94d5fee
```
gftg85: As far as I can see, this commit ID will do the right thing (i.e.: get only revisions __within__ the 3.7 branch).

In `scripts/utilities/update_revision.sh`, the `get_latest_revision` function will run the following command: `git branch -r --contains ${ATSRC_PACKAGE_REV}`. Manually running this on the command-line produces:

```
$ git branch -r --contains 27f6e94d5fee 
  origin/3.7
```

Looks good to me.

PR #814, opened 2019-02-16
Update gcc on AT next

Cherbot:
Bump to revision bf4d9552d7c3


er-1:
ck_binaries test failed because of version mismatch. It expect gcc 9.0.0 whereas it's gcc 9.0.1.

PR #823, opened 2019-02-16
Update valgrind on AT next

Cherbot:
Bump to revision de7fc1a0593c


er-1:
The patch on Makefile.am can't be applied.

PR #838, opened 2019-02-23
Update gcc on AT 12.0

Cherbot:
Bump to revision 269124


er-1:
ck_binairies failed because there is a mismatch between version: expected 8.2.1, got 8.3.1.

PR #843, opened 2019-02-27
Improve fvtr script

er-1:
Fix the script to avoid, for examples, the following lines from the set command:
```
if [ "${MODULES_USE_COMPAT_VERSION:-0}" = '1' ]; then
unset MODULES_USE_COMPAT_VERSION;
MODULES_USE_COMPAT_VERSION=1;
export MODULES_USE_COMPAT_VERSION;
```
This kind of lines produced an awkward output.

Also, remove an old comment from the script with a reference to a file that isn't use any longer.

### DIFF
--- a/fvtr/fvtr.sh
+++ b/fvtr/fvtr.sh
@@ -150,28 +150,15 @@ then
 	exit 1
 fi
 
-# creating a file that will hold the AT_* vars
-# scripts that need to use any of the AT_* vars can
-# access them by calling:
-# array set AT_ENV [ exec cat envvararray.txt ]
-# i.e. $AT_ENV(AT_DEST)
-#
 source $CONFIGFILE
-set > envtempfile.txt
-grep "AT_" envtempfile.txt | sed 's/^> //' | sed "s/'/{/" | \
-	sed "s/'/}/"| sed 's/=/!!!/' > envvars.txt
-while read line
-do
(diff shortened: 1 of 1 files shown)


fvtr/fvtr.sh
```diff
@@ -150,28 +150,15 @@ then
...
-		varval="NOT_SET"
-	fi
-	export $varname="$varval"
-done < envvars.txt
+eval $(
+    set | grep "^AT_" | sed -e "s/^> //" -e "s/'/{/" -e "s/'/}/" -e "s/=/ /" | \
+	while read varname varval
+	do
+	    test -z "$varval" && varval="NOT_SET"
+	    echo "export $varname=\"$varval\""
```
gftg85: Is the `echo` call correct, here?  What effect does it have?

er-1: @gftg85 
In its subshell, `echo` will send to stdout many strings like the following one:
`export AT_VARNAME="A_VALUE"`
The `eval` will evaluate those strings. The variable `AT_VARNAME` will be exported in the current shell (the script one) environment.

If the `export` is done in the while-loop, the variable will be exported in the subshell environment only.

gftg85: Ha!  So that's how it works.  Thanks for the explanation.

PR #860, opened 2019-03-05
Update gcc on AT 12.0

er-1:
Bump to revision 269374 (latest backport made by Bergner)


er-1:
This should be done using Cherbot. Cancelling.

PR #866, opened 2019-03-09
Update glibc on AT next

Cherbot:
Bump to revision 86bdd49d9366


er-1:
The build 1122 overrides the failing build 1118 on SLES12.

PR #889, opened 2019-03-27
Fix ck_lock_elision test

er-1:
The test is ignored if the kernel doesn't support htm-nosc.


tuliom:
The AT 11.0 RHEL7 ppc64le test failure was caused by an intermittent python failure that is unrelated to this PR.

PR #940, opened 2019-04-27
Update mpfr on AT 12.0

Cherbot:
Bump to revision 13478


er-1:
The failure on Ubuntu 18 is expected as it's not supported yet.

PR #948, opened 2019-05-06
Add support of Debian 9 on AT 11.0 and AT 12.0

er-1:


er-1:
AT 12.0 is expecting to fail on Ubuntu18.

PR #955, opened 2019-05-13
Add the package systemtap-sdt-dev to requirements

### LINKED ISSUE #946: missing software-defined tracepoints
A colleague has been trying to use `splat` (https://github.com/open-power-sdk/splat) with an application compiled and linked with Advance Toolchain.  `splat` is a wrapper around the `perf` command which makes use of "software defined tracepoints" (SDTs).  SDTs are predefined tracepoints in user-level/library code which can be enabled and used to do system-wide tracing on user-mode symbols.

Glibc has SDTs defined for many things, including what `splat` uses:  various `pthread_mutex` function calls.  I'm not an expert here, but from what I've found (looking at the code for the `perf` command, the SDT definitions show up in an ELF "note" section in the library.  Some evidence to support that:
```
$ readelf --notes /lib64/libpthread-2.17.so | grep sdt | wc -l
115
```
 
I'm not seeing the same information in the corresponding Advance Toolchain library:
```
$ readelf --notes /opt/at12.0/lib64/libpthread-2.28.so | grep sdt | wc -l
0
```
 
It looks like, in this case, glibc's `configure` script will enable the generation of SDTs by looking for `<sys/sdt.h>` and successfully compiling a small test case with it.  The script ensures that `USE_STAP_PROBE` is defined for the glibc build if that test case works.  `USE_STAP_PROBE` is then used in glibc's `include/stap-probe.h`, which defines `LIBC_PROBE`, which is the macro used to generate these various SDTs, as can be seen in glibc's `nptl/pthread_mutex_init.c`, for example.

I think there is a simple fix, and that is to make sure the build systems have the packages installed that include `sys/sdt.h`.
On Red Hat:  `systemtap-sdt-devel`
On Ubuntu: `systemtap-sdt-dev`
 
I tried to figure out where a dependency should be added, to avoid this in the future, and I think it would be in `configs/*/distros/*`, but I want to confirm because I am not sure.

er-1:
Added the package to requirement on AT 11.0, 12.0 and next.
Enabled systemptap when building glibc.

Fix #946 
    
### DIFF
--- a/fvtr/systemtap/systemtap.exp
+++ b/fvtr/systemtap/systemtap.exp
@@ -1,0 +1,40 @@
+#! /usr/bin/expect
+# Copyright 2019 IBM Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+#
+# Check if Glibc has SDTs defined.
+
+source shared.exp
+
+if { $env(AT_MAJOR_VERSION) < 11.0 } {
+    printit "Skipping: systemtap isn't enabled on version older than AT 11.0"
+    exit $ENOSYS
+}
+set lib "lib64"
+if { $TARGET32 } {
+    set lib "lib"
+}
+if { $env(AT_CROSS_BUILD) == "yes" } {
+    set readelf_exe $env(AT_DEST)/bin/$env(AT_TARGET)-readelf
+} else {
+    set readelf_exe $env(AT_DEST)/bin/readelf
+}
+set sdt [exec $readelf_exe --notes {*}[glob $env(AT_DEST)/$lib/libpthread-*.so] | \
+	     grep stapsdt | wc -l]
+if { $sdt > 0 } {
+    exit 0
+}
+
+exit 1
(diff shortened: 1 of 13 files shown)


er-1:
AT 12.0 is expecting to fail on Ubuntu18.

tuliom:
Is the at9.0 change intentional? I'm asking because you replaced the symlink on at11.0.

er-1:
> Is the at9.0 change intentional?

That's a mistake.

er-1:
> > Is the at9.0 change intentional?

> That's a mistake.

@tuliom it's fixed.

er-1:
@tuliom I added a new testcase.

ThinkOpenly:
I'm able to add the probes I want using Erwan's build (AT-next on Ubuntu 18.04).

Tested-by: Paul A. Clarke <pc@us.ibm.com>

fvtr/systemtap/systemtap.exp
```diff
@@ -0,0 +1,36 @@
...
+    printit "Skipping: systemtap isn't enabled on version older than AT 11.0"
+    exit $ENOSYS
+}
+set lib "lib64"
+if { $TARGET32 } {
+    set lib "lib"
+}
+set readelf_path [exec which readelf]
+set sdt [exec $readelf_path --notes  "$env(AT_DEST)/$lib/libpthread.so" | \
+	     grep sdt | wc -l]
```
tuliom: Don't you need to use `eval` here, e.g. https://github.com/advancetoolchain/advance-toolchain/blob/master/fvtr/shared.exp#L402

gftg85: Also, 'sdt' is a simple string, are you sure this won't match anything else?  Maybe replace it with stapsdt?

er-1: @gftg85 I replaced `sdt` with `stapsdt`.

fvtr/systemtap/systemtap.exp
```diff
@@ -0,0 +1,36 @@
...
+
+if { $env(AT_MAJOR_VERSION) < 11.0 } {
+    printit "Skipping: systemtap isn't enabled on version older than AT 11.0"
+    exit $ENOSYS
+}
+set lib "lib64"
+if { $TARGET32 } {
+    set lib "lib"
+}
+set readelf_path [exec which readelf]
```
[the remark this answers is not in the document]
er-1: @tuliom I updated the code with the referenced one.

er-1:
@gftg85 Good point. I followed your hint and updated the patch.

PR #950, opened 2019-05-11
Update gcc on AT next

Cherbot:
Bump to revision f1f1a94e47cd

PR #967, opened 2019-05-22
Disable SETCONTEXT_CLOBBERS_TLS on AT next

er-1:
GCC has started to generate P9 instructions while building support for Go.
We need to disable SETCONTEXT_CLOBBERS_TLS to avoid error on build.

Close #950 
    

tuliom:
@er-1 Yes, but I suggest to replace `(related with P9 instructions)` with `(avoid running instructions for the optimized cpu at configure time)`

er-1:
@tuliom done.

PR #1006, opened 2019-06-13
Add support of RHEL8 on AT next

### LINKED ISSUE #1005: AT next for RHEL8

er-1:
Fix #1005 

### DIFF
--- a/configs/next/distros/redhat-8.mk
+++ b/configs/next/distros/redhat-8.mk
@@ -1,0 +1,97 @@
+# Copyright 2019 IBM Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Distro dependent basic definitions for RedHat Enterprise Server 8
+# =================================================================
+# This Makefile include contains the definitions for the distro being used
+# for the build process. It contains kernel version infos and distro specific
+# sanity checks.
+
+# Informs the distro supported power archs
+AT_SUPPORTED_ARCHS := ppc64le
+
+# You may force the BUILD_IGNORE_AT_COMPAT general condition by distro
+#BUILD_IGNORE_AT_COMPAT := yes
+
+# Kernel version to build toolchain against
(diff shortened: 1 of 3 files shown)


configs/next/distros/redhat-8.mk
```diff
@@ -0,0 +1,97 @@
...
+# Inform the mainly supported distros
+AT_SUPPORTED_DISTROS := RHEL8
+
+# Inform the compatibility supported distros
+AT_COMPAT_DISTROS :=
+
+# Sign the repository and packages
+AT_SIGN_PKGS := yes
+AT_SIGN_REPO := yes
+AT_SIGN_PKGS_CROSS := no
```
tuliom: Oh! These are still not signed...
We should change this in the future.  Doesn't need to change it for this PR though.

configs/next/distros/redhat-8.mk
```diff
@@ -0,0 +1,97 @@
...
+AT_SIGN_REPO_CROSS := yes
+
+# ID of GNUPG key used to sign our packages (main/compat)
+AT_GPG_KEYID  := 6976A827
+AT_GPG_KEYIDC := 6976A827
+AT_GPG_KEYIDL := 6976A827
+# ID of GNUPG key used to sign our repositories (main/compat)
+AT_GPG_REPO_KEYID  := 6976A827
+AT_GPG_REPO_KEYIDC := 3052930D
```
tuliom: I don't think this is necessary, based on the rest of the changes, but a perfectly correct value would have been 6976A827 too because RHEL 7 was also signed with that key.

configs/next/distros/redhat-8.mk
```diff
@@ -0,0 +1,97 @@
...
+    AT_COMMON_PKGS_REQ := zlib-devel ncurses-devel flex bison texinfo \
+                          createrepo glibc-devel subversion cvs gawk \
+                          rsync curl bc automake libstdc\\+\\+-static \
+                          redhat-lsb-core autoconf bzip2-[0-9] libtool-[0-9] \
+                          gzip rpm-build-[0-9] rpm-sign gcc-c++ imake wget \
+                          docbook2X autoconf-archive libffi-devel python36 \
+                          systemtap-sdt-devel
+    AT_CROSS_PGMS_REQ :=
+    AT_NATIVE_PGMS_REQ :=
+    AT_COMMON_PGMS_REQ := git_1.7 make_4.0
```
tuliom: The `AT_*_PGMS_REQ` variables were used just to document programs that had to be manually built because the packages available in the distro were too old.  I suspect that's the case for git and make.
Do we really need to list them here?  Can't we move them to `AT_COMMON_PKGS_REQ`?

configs/next/distros/redhat-8.mk
er-1: Fixed

configs/next/distros/redhat-8.mk
er-1: Fixed

PR #1017, opened 2019-06-24
Add support of SLES 15 on AT next

er-1:


configs/next/distros/suse-15.mk
```diff
@@ -0,0 +1,91 @@
+# Copyright 2017 IBM Corporation
```
tuliom: 2019

er-1: Fixed

PR #1018, opened 2019-06-28
Drop distros from AT next

er-1:
Drop the following distros from AT next:
- RHEL 7
- SLES 12
- Ubuntu 16.04


tuliom:
I know it's unrelated to this patch, but we should have one Debian file in this directory too.
At least a symlink to configs/12.0/distros/debian-9.mk

er-1:
> I know it's unrelated to this patch, but we should have one Debian file in this directory too.
> At least a symlink to configs/12.0/distros/debian-9.mk

My goal is to add Debian 10 in the next weeks. Debian 10 is to be released on July 6th.

PR #1044, opened 2019-07-17
Fix cross build of glibc

er-1:
When building glibc, the systemtap headers are copied to a directory where gcc looks at. After the build the systemtap headers are removed.
When cross building glibc, 64 bits version and 32 version can be built at the same time. Meaning a build can copy and remove the systemtap headers while the other build is compiling ; the systemtap headers will be missing to complete the second build.

To avoid this, the builds can't be done in parallel. The 64 bits build will occur before the 32 bits build.

### DIFF
--- a/configs/11.0/packages/glibc/glibc.mk
+++ b/configs/11.0/packages/glibc/glibc.mk
@@ -1,1 +1,68 @@
-../../../10.0/packages/glibc/glibc.mk
+# Copyright 2017 IBM Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


tuliom:
I don't think this is the right way to solve the issue.
It's going to decrease parallelism and increase build time.
Let me find a suggestion....

er-1:
> I don't think this is the right way to solve the issue.
I agree.

tuliom:
I proposed an alternative solution in PR #1045 .

er-1:
PR #1045 is a far better fix.

PR #1043, opened 2019-07-16
Excludes even more python tests in the FVTR

tuliom:
In order for the FVTR to work, we need reliable tests.
Unfortunately, some of the python tests have been failing
intermittently, making it difficult to trust the entire python
FVTR test.

In order to improve this situation, this patch excludes the following
python tests from the FVTR:
 - test_faulthandler
 - test_imaplib
 - test_selectors
 - test_time

Issue #966, opened 2019-05-21
gcc failed to build on Debian 9


er-1:
On stage 2:
```
/usr/bin/g++ -std=gnu++98 -c   -isystem /home/tcbot/bot-slave/at_opt/at11.0/include -DIN_GCC     -fno-exceptions -fno-rtti -fasynchronous-unwind-tables -W -Wall -Wno-narrowing -Wwrite-strings -Wcast-qual -Wno-format -Wmissing-format-attribute -Woverloaded-virtual -pedantic -Wno-long-long -Wno-variadic-macros -Wno-overlength-strings -fno-common  -DHAVE_CONFIG_H -DGENERATOR_FILE -fno-PIE -I. -Ibuild -I/home/tcbot/bot-slave/at/src/at11.0-4.debian-9_ppc64le_ppc64le/sources/gcc/gcc -I/home/tcbot/bot-slave/at/src/at11.0-4.debian-9_ppc64le_ppc64le/sources/gcc/gcc/build -I/home/tcbot/bot-slave/at/src/at11.0-4.debian-9_ppc64le_ppc64le/sources/gcc/gcc/../include  -I/home/tcbot/bot-slave/at/src/at11.0-4.debian-9_ppc64le_ppc64le/sources/gcc/gcc/../libcpp/include  \
        -o build/genrecog.o /home/tcbot/bot-slave/at/src/at11.0-4.debian-9_ppc64le_ppc64le/sources/gcc/gcc/genrecog.c
In file included from /usr/include/c++/6/cstdlib:75:0,
                 from /usr/include/c++/6/bits/stl_algo.h:59,
                 from /usr/include/c++/6/algorithm:62,
                 from /home/tcbot/bot-slave/at/src/at11.0-4.debian-9_ppc64le_ppc64le/sources/gcc/gcc/system.h:221,
                 from /home/tcbot/bot-slave/at/src/at11.0-4.debian-9_ppc64le_ppc64le/sources/gcc/gcc/genrecog.c:109:
/usr/include/stdlib.h:56:1: error: ‘__BEGIN_NAMESPACE_STD’ does not name a type
 __BEGIN_NAMESPACE_STD
 ^~~~~~~~~~~~~~~~~~~~~
```

er-1:
This blocks PR #963.

er-1:
This blocks PR #977.

PR #1045, opened 2019-07-18
Refactor code for enabling systemtap probes in glibc

tuliom:
Replaces the previous code that enabled systemtap probes in glibc with a
new one that copies the header files to the build directory.
This has the advantage of allowing 2 builds (e.g. 32 and 64 bits) to
execute at the same time because each build has its own copy of the
header files.  Its removal will not break the other build.

This is also refactoring the FVTR testcase in order to improve the log
output and to guarantee the test will check both 32 and 64 bit libraries
when necessary.

Related to issue #946.


er-1:
The failures on Debian9 is expected (cf #966 ).
The others failures are due to failing python tests. This will be fixed by PR #1043 .

PR #1104, opened 2019-08-29
Allow the installation tool to install AT next packages

er-1:
When building AT next, the packages have the next version of AT in their name,
not "next". So, when the installation tool installs those packages it fails
because of this mistmatch.
The installation tool has been updated to find the AT next packages.

This is part of a request to build AT next on a weekly basis.


er-1:
> The commit message is missing your signed-off-by line.

My bad...
It's fixed.

PR #1122, opened 2019-09-10
Update the documentation

er-1:
Add information about AT 13.0 in the website.
Also, improve the use of the HTML tag `pre`.

### DIFF
--- a/docs/advtool-installation.html
+++ b/docs/advtool-installation.html
@@ -47,6 +47,12 @@
 
 
 <!-- Begin content -->
+<style>
+pre {
+    margin-top: 1em;
+    margin-left: 4em;
+}
+</style>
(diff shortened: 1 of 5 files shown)


docs/advtool-installation.html
```diff
@@ -561,7 +637,7 @@ <h4 class="ibm-h4">Taking advantage of leading edge tools and libraries</h4>
     </li>
   </ol>
 
-  <p>APT supports package upgrades for new revision releases (i.e. 12.0-0 to 12.0-1). For new major releases (i.e. 11.0-3 to 12.0-0), please proceed as in a normal installation.</p>
+  <p>APT supports package upgrades for new revision releases (i.e. 13.0-0 to 13.0-1). For new major releases (i.e. 12.0-1 to 13.0-0), please proceed as in a normal installation.</p>
```
rzinsly: I'm ok with this change but I don't think you need to update this example every new major release.

PR #1149, opened 2019-10-09
Fix: Mixing AT and host env during gcc build stages

### LINKED ISSUE #966: gcc failed to build on Debian 9

rff:
On gcc_2 (and similarly with gcc_3), it should use gcc_1, binutils_1 and
glibc_1 but it was using host gcc and because of that mixing at and host
includes and libs. Also it was mixing running binaries compiled against
AT with runtime from host, using host loader and host libs (instead of
at versions).

This bug not always present itself if the host and AT libs versions are
sufficiently compatible but presented conflicts in some combinations,
notably AT 11/12 with Debian 9 and (only changing CC/CXX) AT 13 and
Ubuntu 18.04.

Fix #966 (gcc failed to build on Debian 9)

Signed-off-by: Raoni Fassina Firmino <raoni@linux.ibm.com>


configs/11.0/packages/gcc/stage_2
```diff
@@ -109,7 +110,8 @@ atcfg_configure() {
 			--with-mpc-include="${at_dest}/include" \
 			--with-mpc-lib=${libdir} \
 			--with-as="${at_dest}/bin/as" \
-			--with-ld="${at_dest}/bin/ld"
+			--with-ld="${at_dest}/bin/ld" \
+			--with-stage1-ldflags="-Wl,--dynamic-linker=${at_dest}/lib64/ld64.so.2"
```
rff: This line is the trick one and that I am not sure is the right approach.

By setting CC/CXX it fix all search paths for include and linking while building gcc_2 and gcc_3.
But It does not fix the library search path for execution. This is clear during gcc_2 compilation. It compiles a helper program called genautomata and tries to run it as parte of the make script. It is built against glibc from AT but it is run with the host loader (/lib64/ld.so.2) which uses the host library paths to search libs. In the case of AT13 and ubuntu 18.04 it fails because the host AT is too old for the newly build program, and the loader complains.

There is some ways to fix it. The ones I though of are:
a) Use rpath to hardcode runtime library paths for built binaries.
b) Use LD_LIBRARY_PATH to use AT libraries.
c) Change the `.interp` loader in the elf of built binaries for the AT ld.so.2 that is already configured with the right search paths.

I choose (c). It seems that gcc by default (or ld) uses the same loader as itself for the elfs it build (That is why genautomata built with gcc_1 have the host loader in `.interp`). But after creating glibc_1 we can start using the new loader. And once you change it for gcc_2 it becomes the default loader for all elfs built with it for now on and later gcc_3 and so on.

configs/11.0/packages/gcc/stage_2
```diff
@@ -72,8 +72,9 @@ atcfg_pre_configure() {
 atcfg_configure() {
 	if [[ "${cross_build}" == "no" ]]; then
 		# Configure command for native builds
-		CC="${cc_64}" \
-		CXX="${cxx_64}" \
```
rzinsly: You could clean the atcfg_pre_configure() to just set ${libdir} since we won't need ${c*_64} anymore.

configs/11.0/packages/gcc/stage_2
rzinsly: I'm ok with your solution.
We set LIBRARY_PATH in the stage 4 for AT 9.0, you could do the same to have a sort of consistency if you don't need to set it again on the later stages. I wasn't expect this behaviour to happen since you are setting CC/CXX to AT.

configs/11.0/packages/gcc/stage_2
rff: Done. gave a general cleanup on atcfg_pre_configure for stage 2 and 3 and removed the now unecessary ${c*_64} variables.

configs/13.0/packages/gcc/stage_4
```diff
@@ -38,13 +38,16 @@ atcfg_configure() {
 		# Configure command for native builds
 		PATH=${at_dest}/bin:${PATH} \
 		CC="${at_dest}/bin/gcc" \
+		CXX="${at_dest}/bin/g++" \
+		AS="${at_dest}/bin/as" \
+		LD="${at_dest}/bin/ld" \
+		AR="${at_dest}/bin/ar" \
+		NM="${at_dest}/bin/nm" \
```
gftg85: For consistency's sake, could you also add RANLIB here, as you did for versions 11.0 and 12.0?

configs/11.0/packages/gcc/stage_2
```diff
@@ -159,9 +156,7 @@ atcfg_configure() {
 
 # Make build command
 atcfg_make() {
-	# With -isystem we make sure that system's gcc (stage1) will recognize AT
-	# headers as system headers and give them special treatment.
-	${SUB_MAKE} STAGE1_CFLAGS="-isystem ${at_dest}/include" STAGE1_CXXFLAGS="-isystem ${at_dest}/include" all
```
gftg85: I never knew what this accomplished and I'm glad that it's going away.  Thanks.

configs/11.0/packages/gcc/stage_2
gftg85: >By setting CC/CXX it fix all search paths for include and linking while building gcc_2 and gcc_3.
But It does not fix the library search path for execution. This is clear during gcc_2 compilation. It compiles a helper program called genautomata and tries to run it as part of the make script. It is built against glibc from AT but it is run with the host loader (/lib64/ld.so.2) which uses the host library paths to search libs. In the case of AT13 and ubuntu 18.04 it fails because the host glibc is too old for the newly build program, and the loader complains.

I think you should add this paragraph **as is** to the commit message.  And I suggest that you add it as a comment in the code, with a little rephrasing of the first 3 sentences so that they don't describe what's changed, but what's accomplished with the code.  Suggestion:

_The use of --with-stage1-ldflags is required during gcc_2 compilation. gcc_2 compiles a helper program called genautomata and tries to run it as part of the make script. It is correctly built against glibc from AT, but it is run with the host loader (/lib64/ld.so.2), which uses the host library paths to search libs, potentially leading to build failures._

(as a comment in the code).

rff:
Sorry for a new update after the patch has already being cleared.
Changes in this version:
1) Incorporated all sugestions from @gftg85 
2) Rebase on top of upstream master
3) Added the same changes to AT 10.0 gcc_2. Because I already had changed stage_optimized that is shared from 10 to 12 I decided to fix stages 2, 3 and 4 for AT 10.
4) Added an extra setup, updating PATH when calling make in atcfg_make() for stages 2 and 3, the same way it was already being setup on stages 4 and optimized.
5) Updated commit message.

tuliom:
This is weird.  The glibc build failed because the kernel headers are not greater or equal v4.9.
Let me restart the build....

rff:
> This is weird. The glibc build failed because the kernel headers are not greater or equal v4.9.
> Let me restart the build....

Indeed. And was a glibc_1 that failed, and it uses gcc_1. This patch didn't touch gcc_1.
I started a new build manually and it is going OK, already on gcc_3, so it passed the previous failing point.

configs/11.0/packages/gcc/stage_2
tuliom: @rff Oops.  The loader path is correct correct for ppc64le (ELFv2), but is wrong for ppc and ppc64.  In those cases they should point to:
- ppc: `lib/ld.so.1`
- ppc64: `lib64/ld64.so.1`

PR #1160, opened 2019-10-18
Fix gcc_2 build with wrong loader name for ppc and ppc64

rff:
This commit fixes a bug introduzed in:

    commit: 2273ba38d8a493018e3ab63dc96ea7376e2cd3d0
    subject: Fix: Mixing AT and host env during gcc build stages

Whilst explicitly setting the loader the linker should set for generated
ELFs, I was used the  wrong path and name for the loader in ppc and
ppc64 arquitectures (so, only working on ppc64le).

Signed-off-by: Raoni Fassina Firmino <raoni@linux.ibm.com>

-----------------------------------------------------------------------------------------

More about the bug/fix here:
https://github.com/advancetoolchain/advance-toolchain/pull/1149#discussion_r336489887

### DIFF
--- a/configs/10.0/packages/gcc/stage_2
+++ b/configs/10.0/packages/gcc/stage_2
@@ -54,6 +54,13 @@ atcfg_pre_configure() {
 	if [[ "${cross_build}" == "no" && "${build_arch}" == ppc64* ]]; then
 		libdir="${at_dest}/lib64"
 	fi
+	if [[ "${cross_build}" == "no" ]]; then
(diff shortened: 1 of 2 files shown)


rff:
Needed to push a v2 because there was this snaky error in my bash:
```diff
-       if [[ "${cross_build}" == "no"]]; then
+       if [[ "${cross_build}" == "no" ]]; then
```
Can you see it? I did not, so I made a broken untested pull-request. Sorry about that :(

PR #1223, opened 2019-12-03
Fix display of gcc -v on AT >= 11.0

### LINKED ISSUE #738: Validate the contents of gcc/REVISION
When fixing issue #497 we added code to validate and transform the contents of gcc/REVISION.

However, there are still issues in AT 12.0.

We need to add a stricter validation of the format before transforming it.
Builds must fail if they do not pass this validation.

This issue also affects AT 10.0.

er-1:
At stage 1 of gcc, it checks we are using an IBM branch (if not it ends the build) and then updates gcc/REVISION as shown in the following example:
from: `[ibm/gcc-9-branch revision 277354]`
to: `[ibm/gcc-9-branch revision 277364, based upon FSF revision 277354]`

Removed unreacheable code from AT next stage 1.

Close #738

### DIFF
--- a/configs/11.0/packages/gcc/stage_1
+++ b/configs/11.0/packages/gcc/stage_1
@@ -26,30 +26,28 @@ ATCFG_BUILD_STAGE_T='dir'
 
 # Required pre build hacks
 atcfg_pre_hacks() {
+	# IBM maintains branches on the svn repository for GCC.
+	# Check we are using an IBM branch. If not, stop everything.
+	if [[ ! -s ${ATSRC_PACKAGE_WORK}/gcc/REVISION ]]; then
+		echo "ERROR: gcc/REVISION is missing and/or empty"
+		exit 1
+	fi
+	if ! grep -Eqs "^\[ibm/gcc-[0-9]+-branch revision [0-9]+]$" ${ATSRC_PACKAGE_WORK}/gcc/REVISION; then
+		echo "ERROR: Not using an IBM branch"
+		exit 1
+	fi
 	# Overwrite the existing DEV-PHASE with Advance-Toolchain-XX
 	mv ${ATSRC_PACKAGE_WORK}/gcc/DEV-PHASE ${ATSRC_PACKAGE_WORK}/gcc/DEV-PHASE.back
 	echo "Advance-Toolchain-${at_major_internal}" > ${ATSRC_PACKAGE_WORK}/gcc/DEV-PHASE
 	# Expand REVISION with the requested revision
-	if [ -s ${ATSRC_PACKAGE_WORK}/gcc/REVISION]; then
-		cp ${ATSRC_PACKAGE_WORK}/gcc/REVISION ${ATSRC_PACKAGE_WORK}/gcc/REVISION.back
-		sed -i "s/ \([^ ][^ ]*\)]$/ ${ATSRC_PACKAGE_REV}, based upon FSF revision \1]/" \
-			${ATSRC_PACKAGE_WORK}/gcc/REVISION
-		cmp --silent ${ATSRC_PACKAGE_WORK}/gcc/REVISION.back ${ATSRC_PACKAGE_WORK}/gcc/REVISION || exit 1
-	else
-		echo "[revision ${ATSRC_PACKAGE_REV}]" > ${ATSRC_PACKAGE_WORK}/gcc/REVISION
-	fi
-	# IBM maintains branches on the svn repository for GCC.  Such branches
-	# do not update the release timestamp (gcc/DATESTAMP).  The following
-	# lines patch the release timestamp with the date of latest entry in
-	# the IBM branch ChangeLog, if the ChangeLog exists.  If it doesn't
-	# exist, then we are probably not using an IBM branch.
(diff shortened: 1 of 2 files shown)


configs/11.0/packages/gcc/stage_1
```diff
@@ -26,30 +26,28 @@ ATCFG_BUILD_STAGE_T='dir'
...
+	# IBM maintains branches on the svn repository for GCC.
+	# Check we are using an IBM branch. If not, stop everything.
+	if [[ ! -s ${ATSRC_PACKAGE_WORK}/gcc/REVISION ]]; then
+		echo "ERROR: gcc/REVISION is missing and/or empty"
+		exit 1
+	fi
+	if ! grep -Eqs "^\[ibm/gcc-[0-9]+-branch revision [0-9]+]$" ${ATSRC_PACKAGE_WORK}/gcc/REVISION; then
+		echo "ERROR: Not using an IBM branch"
+		exit 1
+	fi
```
gftg85: OK.  I like this check better than what we have today (the comparison between `REVISION.back` and the `sed` output)

configs/11.0/packages/gcc/stage_1
```diff
@@ -26,30 +26,28 @@ ATCFG_BUILD_STAGE_T='dir'
...
-	# the IBM branch ChangeLog, if the ChangeLog exists.  If it doesn't
-	# exist, then we are probably not using an IBM branch.
-	if [ -f ${ATSRC_PACKAGE_WORK}/gcc/ChangeLog.ibm ]; then
-		mv ${ATSRC_PACKAGE_WORK}/gcc/DATESTAMP ${ATSRC_PACKAGE_WORK}/gcc/DATESTAMP.back
-		echo "$(head -n 1 ${ATSRC_PACKAGE_WORK}/gcc/ChangeLog.ibm | cut -f 1 -d " " | sed "s/-//g")" \
-			  > ${ATSRC_PACKAGE_WORK}/gcc/DATESTAMP
-	else
-		echo "WARNING: Not using an IBM branch"
-	fi
+	cp ${ATSRC_PACKAGE_WORK}/gcc/REVISION ${ATSRC_PACKAGE_WORK}/gcc/REVISION.back
```
gftg85: The copy to REVISION.back is no longer necessary and could be removed.

configs/11.0/packages/gcc/stage_1
```diff
@@ -26,30 +26,28 @@ ATCFG_BUILD_STAGE_T='dir'
...
+	sed -i "s/ \([^ ][^ ]*\)]$/ ${ATSRC_PACKAGE_REV}, based upon FSF revision \1]/" \
+		${ATSRC_PACKAGE_WORK}/gcc/REVISION
+	# The following lines patch the release timestamp (gcc/DATESTAMP)
+	# with the date of latest entry in the IBM branch ChangeLog.
+	mv ${ATSRC_PACKAGE_WORK}/gcc/DATESTAMP ${ATSRC_PACKAGE_WORK}/gcc/DATESTAMP.back
```
gftg85: Has the copy to DATESTAMP.back ever been neessary?

configs/next/packages/gcc/stage_1
```diff
@@ -34,7 +34,6 @@ atcfg_pre_hacks() {
 		cp ${ATSRC_PACKAGE_WORK}/gcc/REVISION ${ATSRC_PACKAGE_WORK}/gcc/REVISION.back
 		sed -i "s/ \([^ ][^ ]*\)]$/ ${ATSRC_PACKAGE_REV}, based upon FSF revision \1]/" \
 			${ATSRC_PACKAGE_WORK}/gcc/REVISION
-		cmp --silent ${ATSRC_PACKAGE_WORK}/gcc/REVISION.back ${ATSRC_PACKAGE_WORK}/gcc/REVISION || exit 1
```
gftg85: The master branch of GCC does not have this file (gcc/REVISION), so I'm wondering what problem is this hunk trying to fix.  Is there a problem, at all?

Maybe the whole if-else block should be replaced with:

```
if [ ! -s ${ATSRC_PACKAGE_WORK}/gcc/REVISION]; then
  echo "[revision ${ATSRC_PACKAGE_REV}]" > ${ATSRC_PACKAGE_WORK}/gcc/REVISION
fi
```
(notice the inclusion of the `!` in the if statement)

configs/11.0/packages/gcc/stage_1
er-1: It could be useful for debugging. I've always thought we keep a .back file for that.

configs/11.0/packages/gcc/stage_1
er-1: It could be useful for debugging. I've always thought we keep a .back file for that.

configs/11.0/packages/gcc/stage_1
gftg85: OK.

configs/11.0/packages/gcc/stage_1
gftg85: OK.

PR #1252, opened 2020-01-11
Update gcc on AT next

Cherbot:
Bump to revision d3cf980217ce


er-1:
It failed because GCC has been migrated to GIT. We need to update the URL to correct repository.

PR #1256, opened 2020-01-11
Update valgrind on AT next

Cherbot:
Bump to revision 3e48ab0803d4

PR #1257, opened 2020-01-18
Update valgrind on AT next

Cherbot:
Bump to revision f12114be2391


er-1:
It failed while building valgrind:
```
it_main.c: In function ‘vgItrace_post_clo_init’:
it_main.c:353:22: error: ‘VexControl’ has no member named ‘guest_chase_thresh’
  353 |  VG_(clo_vex_control).guest_chase_thresh = 0;
      |                      ^
```

mscastanho:
Isn't this the same issue as in #1256? So the fix proposed didn't work?

er-1:
> Isn't this the same issue as in #1256? So the fix proposed didn't work?

Same issue.
Either the fix doesn't work or it hasn't been taken in account upstream yet.

er-1:
I'm closing this PR to check if the issue has been fixed upstream.

PR #1259, opened 2020-01-20
Remove flag to ignore warning on glibc

mscastanho:
The issue causing a maybe-uninitialized warning on glibc builds
on AT next has been fixed upstream by commit 9f8b135f76ac794.

We should not need to ignore this warning anymore to successfully
build AT next.


mscastanho:
Actually, this should only be merged once glibc is updated to revision 9f8b135f76ac794 or newer on AT next.

er-1:
> Actually, this should only be merged once glibc is updated to revision 9f8b135f76ac794 or newer on AT next.

I think you can update the revision in your patch.

mscastanho:
Done.

mscastanho:
Just rebased against the current master. I kept the update to glibc sources as the recent update still did not include `9f8b135f76ac794`.

PR #1262, opened 2020-01-27
Update GCC source repository

er-1:
Since January 13th, gcc has moved to git from svn.

Because of the way to 'hide' vendors branches, it needs to update AT < next.
gcc/sources files include git commands to reach IBM branches. For that,
ATSRC_PACKAGE_GIT can be a list of git commands.
It also needs to update the update_revision script to be able to check if
there is an update in IBM branches.

All revisions in gcc/sources files have been updated to match the commit id
in the new gcc repository.

Close #1252 


scripts/utilities/update_revision.sh
```diff
@@ -150,7 +160,13 @@ get_latest_revision ()
 		if [[ $(expr index "${branch}" "HEAD") -ne 0 ]]; then
 			branch="HEAD"
 		else
+		    # Since GCC moved to git (January 2020), the name of the branch
+		    # has looked like: ibm/<branch>
+		    if [[ $package == "gcc" ]] && [[ $configset != "next" ]]; then
```
mscastanho: Why is it necessary to exclude AT-next here?

scripts/utilities/update_revision.sh
```diff
@@ -476,26 +492,27 @@ update_revision ()
 	print_msg 0 "Update complete.";
 }
 
+package=$(readlink -f $1 | tr "/" "\n" | tail -n 2 | head -n 1)
+configset=$(readlink -f $1 | tr "/" "\n" | tail -n 4 | head -n 1)
```
mscastanho: Works as expected.

scripts/utilities/update_revision.sh
```diff
@@ -121,14 +121,18 @@ print_msg ()
 #
 # Parameters:
 #    $1 - sources config path
+#    $2 - package
+#    $3 - configset
```
mscastanho: Maybe also update the description of `$1`, as it expects the repo's URL, instead of sources config path.

configs/11.0/packages/gcc/sources
```diff
@@ -32,9 +32,11 @@ ATSRC_PACKAGE_RELFIXES=
 ATSRC_PACKAGE_NAMESUFFIX="[C, C++ (g++), fortran, Go]"
 ATSRC_PACKAGE_STR_VER="${ATSRC_PACKAGE_NAME} ${ATSRC_PACKAGE_VER}-ibm-r${ATSRC_PACKAGE_REV}"
 ATSRC_PACKAGE_PRE="test -d gcc-${ATSRC_PACKAGE_VER}-${ATSRC_PACKAGE_REV}"
-ATSRC_PACKAGE_CO=([0]="svn co --revision ${ATSRC_PACKAGE_REV} svn://gcc.gnu.org/svn/gcc/branches/ibm/gcc-7-branch" \
-		  [1]="svn co --revision ${ATSRC_PACKAGE_REV} http://gcc.gnu.org/svn/gcc/branches/ibm/gcc-7-branch")
-ATSRC_PACKAGE_POST="mv gcc-7-branch gcc-${ATSRC_PACKAGE_VER}-${ATSRC_PACKAGE_REV}"
+ATSRC_PACKAGE_CO=([0]="git clone git://gcc.gnu.org/git/gcc.git")
+ATSRC_PACKAGE_GIT=([0]="git config --local remote.origin.fetch +refs/vendors/ibm/heads/*:refs/remotes/origin/ibm/*" \
+                   [1]="git fetch origin" \
+                   [2]="git checkout -b gcc-${ATSRC_PACKAGE_VER}${ATSRC_PACKAGE_REV:+-${ATSRC_PACKAGE_REV}} ${ATSRC_PACKAGE_REV}")
```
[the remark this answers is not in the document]
tuliom: @mscastanho In the past, it had to change the values frequently for tests (including removing the value of the revision) and we did not want to rewrite the entire file.
So, we ended up writing all of them in way that would allow to remove the revision and continue working.

configs/11.0/packages/gcc/sources
```diff
@@ -32,9 +32,11 @@ ATSRC_PACKAGE_RELFIXES=
 ATSRC_PACKAGE_NAMESUFFIX="[C, C++ (g++), fortran, Go]"
 ATSRC_PACKAGE_STR_VER="${ATSRC_PACKAGE_NAME} ${ATSRC_PACKAGE_VER}-ibm-r${ATSRC_PACKAGE_REV}"
 ATSRC_PACKAGE_PRE="test -d gcc-${ATSRC_PACKAGE_VER}-${ATSRC_PACKAGE_REV}"
-ATSRC_PACKAGE_CO=([0]="svn co --revision ${ATSRC_PACKAGE_REV} svn://gcc.gnu.org/svn/gcc/branches/ibm/gcc-7-branch" \
-		  [1]="svn co --revision ${ATSRC_PACKAGE_REV} http://gcc.gnu.org/svn/gcc/branches/ibm/gcc-7-branch")
-ATSRC_PACKAGE_POST="mv gcc-7-branch gcc-${ATSRC_PACKAGE_VER}-${ATSRC_PACKAGE_REV}"
+ATSRC_PACKAGE_CO=([0]="git clone git://gcc.gnu.org/git/gcc.git")
```
tuliom: You could still implement this part with just 2 commands (git-clone + git-checkout):
```
git clone -c remote.origin.fetch="+refs/vendors/ibm/heads/*:refs/remotes/origin/ibm/*" git://gcc.gnu.org/git/gcc.git
git checkout -b gcc-${ATSRC_PACKAGE_VER}${ATSRC_PACKAGE_REV:+-${ATSRC_PACKAGE_REV}} ${ATSRC_PACKAGE_REV}
```
Anyway, your implementation looks good to me.

er-1: > You could still implement this part with just 2 commands (git-clone + git-checkout):

This works with relatively new git. In older git, there was an issue with -c option of clone, especially when setting remote.origin.fetch. That's why I've implemented this part with 3 commands.

scripts/utilities/update_revision.sh
er-1: AT next is feeded with the master branch, not an IBM branch.

configs/11.0/packages/gcc/sources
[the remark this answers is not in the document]
mscastanho: @tuliom Ok then. But in that case, shouldn't both strings be declared that? If we remove the revision, `gcc-${ATSRC_PACKAGE_REV:+-${ATSRC_PACKAGE_REV}}` will evaluate to something like `gcc-9`, while `gcc-${ATSRC_PACKAGE_VER}-${ATSRC_PACKAGE_REV}`, used to rename the directory in the next line, will be `gcc-9-`.

scripts/utilities/update_revision.sh
mscastanho: Ok, thanks for the clarification =)

configs/11.0/packages/gcc/sources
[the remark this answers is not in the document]
er-1: @mscastanho I agree we could rework all `configs/*/packages/*/sources` files to have something lighter and homogeneous, but it's not the goal of the current patch ;)

PR #1264, opened 2020-01-28
Fix AT >= 13.0 builds on RHEL8

mscastanho:
The `fixincludes` step performed during GCC compilation is currently fixing system headers from `/usr/include`, rather than from `<at-build-dir>/opt/*`. This can cause problems due to a mismatch between header versions installed in the system and the ones used by AT. The commit messages have more detail about it.

This problem is likely due to some missing/wrong GCC configuration or by the way `--with-advance-toolchain` flag works. I'm still investigating the root cause.

Meanwhile, I'm adding a hotfix to just skip the specific fix from `fixincludes` that causes the problem (see commit msg to understand why this is OK).

I'm also removing the requirement for `qt5-devel` on RHEL8, as it is not necessary.

### DIFF
--- a/configs/13.0/distros/redhat-8.mk
+++ b/configs/13.0/distros/redhat-8.mk
@@ -71,7 +71,13 @@ endef
 # Inform the list of packages to check
 ifndef AT_DISTRO_REQ_PKGS
     AT_CROSS_PKGS_REQ :=
-    AT_NATIVE_PKGS_REQ := libxslt docbook-style-xsl qt5-devel \
+    # NOTE: qt5 devel packages are also required. A `qt5-devel` meta
+    # package is provided by RHEL8, which has all needed packages listed
+    # as dependencies. However, it is temporarily unavailable on the repos
+    # configured on our build server, so we are TEMPORARILY removing this
+    # dependecy from AT_NATIVE_PKGS_REQ. It should be re-added as soon as
+    # it can be installed to the build server again.
+    AT_NATIVE_PKGS_REQ := libxslt docbook-style-xsl \
                           autogen-libopts sqlite-devel
     AT_COMMON_PKGS_REQ := zlib-devel ncurses-devel flex bison texinfo \
                           createrepo glibc-devel subversion cvs gawk \
--- a/configs/13.0/packages/gcc/sources
+++ b/configs/13.0/packages/gcc/sources
@@ -64,6 +64,11 @@ atsrc_get_patches ()
 		'https://raw.githubusercontent.com/powertechpreview/powertechpreview/master/GCC%20libstdc%2B%2B%20PowerPC%20Patches/9/0001-libstdc-Prevent-LD_LIBRARY_PATH-from-leaking-to-msgf.patch' \
 		9cba9a9b94ca9c1b8958c0d722a9c06b || return ${?}
 
+	# Skip machine_name fix on GCC fixincludes step.
+	at_get_patch \
+		'https://raw.githubusercontent.com/powertechpreview/powertechpreview/fb462137df68be149d94b1cbde1dd141f32361fa/GCC%20PowerPC%20Backport/9/0001-Skip-machine_name-fix-for-powerpc-on-fixincludes-ste.patch' \
+		7b6e39981c83b5f4634693034c7ffbc7 || return ${?}
+
 	return 0
 }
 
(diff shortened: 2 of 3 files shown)


configs/13.0/distros/redhat-8.mk
```diff
@@ -71,7 +71,7 @@ endef
 # Inform the list of packages to check
 ifndef AT_DISTRO_REQ_PKGS
     AT_CROSS_PKGS_REQ :=
-    AT_NATIVE_PKGS_REQ := libxslt docbook-style-xsl qt5-devel \
+    AT_NATIVE_PKGS_REQ := libxslt docbook-style-xsl \
```
tuliom: Why is this being removed?  It used to be important in order to enable features for one of the packages we distribute. I think it was Valgrind...
In other words: I think the commit message needs to clarify why this can be removed now.

configs/13.0/packages/gcc/stage_optimized
```diff
@@ -36,6 +36,20 @@ atcfg_pre_configure() {
...
+    #   ${GCC_SRC}/fixincludes/fixinc.in line 135
+    # the build should work properly without it, so we can skip it
+    # in order to avoid further surprises.
+    case_powerpc='\
+    powerpc*-*-linux-*)\
+        test -f ${MACRO_LIST} &&  echo > ${MACRO_LIST}\
+        ;;'
+    sed -i -e "/\*-\*-vxworks\*)/ i ${case_powerpc}" \
+        "${ATSRC_PACKAGE_WORK}/fixincludes/fixinc.in"
+}
```
tuliom: Isn't it easier to apply a patch to GCC?

configs/13.0/distros/redhat-8.mk
tuliom: Boost still depends on Qt in order to provide the support for compute/interop: https://www.boost.org/doc/libs/1_72_0/libs/compute/doc/html/boost_compute/interop.html

Are the Qt-related files created even without qt5-devel?
They're usually under `include/boost/compute/interop/qt/`

mscastanho: `qt5-devel` was a meta-packaged that is not available anymore on RHEL8. So requiring it to be installed breaks the build. It had all devel-related packages listed as dependencies. qt-related files are being generated all fine, as those dependencies are installed in the system. 

However we may have to find a different way to express qt5 dependencies to append to `AT_NATIVE_PKGS_REQ`

tuliom: @mscastanho According to the [RHEL 8 documentation](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/considerations_in_adopting_rhel_8/changes-to-packages_considerations-in-adopting-rhel-8#new-packages-in-RHEL-8-0_changes-to-packages), qt5-devel should still be available in the [CRB repository](https://developers.redhat.com/blog/2018/11/15/introducing-codeready-linux-builder/).

Maybe this server does not have the CRB repository enabled?

In case it does and the qt5-devel package was indeed removed, this is the list of packages that it used to require:
```
qt5-designer
qt5-linguist
qt5-qdoc
qt5-qhelpgenerator
qt5-qt3d-devel
qt5-qtbase-devel
qt5-qtbase-static
qt5-qtconnectivity-devel
qt5-qtdeclarative-devel
qt5-qtdeclarative-static
qt5-qtlocation-devel
qt5-qtmultimedia-devel
qt5-qtscript-devel
qt5-qtsensors-devel
qt5-qtserialport-devel
qt5-qtsvg-devel
qt5-qttools-devel
qt5-qttools-static
qt5-qtwayland-devel
qt5-qtwebchannel-devel
qt5-qtwebsockets-devel
qt5-qtx11extras-devel
qt5-qtxmlpatterns-devel
qt5-rpm-macros
```

I don't think we need all of them though.

configs/13.0/packages/gcc/stage_optimized
mscastanho: @tuliom True. I did it this way as there is already another sed-based hack there. 

How should I proceed with the patch? Just add the `.patch` to `[...]/packages/gcc` dir and apply it inside `gcc/sources:atsrc_apply_patches()`?

tuliom: @mscastanho No.  We use another repository for that: https://github.com/powertechpreview/powertechpreview

Here is an example on how we do that:
https://github.com/powertechpreview/powertechpreview/commit/9b646544919d52be95d0190d13187d6ce872ae4d

And in the Advance Toolchain code:
https://github.com/advancetoolchain/advance-toolchain/commit/18fe1b4c5d4072e5801d5d4a7a9bfc1de9874acf

mscastanho:
Updated the PR to apply a patch to GCC instead of the sed-based modification.

mscastanho:
@tuliom I just added a comment right next to the line modified and created a related issue (#1270).

PR #1263, opened 2020-01-27
Update valgrind on AT next

Cherbot:
Bump to revision 2052837a3e0b

PR #1265, opened 2020-01-29
Update valgrind on AT next

er-1:
Bump revision to 2052837a3e0b
Also apply a fix on itrace to fix the initialization of guest_chase variable.

Close #1263 

### DIFF
--- a/configs/next/packages/valgrind/sources
+++ b/configs/next/packages/valgrind/sources
@@ -21,7 +21,7 @@
 
 ATSRC_PACKAGE_NAME="Valgrind"
 ATSRC_PACKAGE_VER=3.15.0
-ATSRC_PACKAGE_REV=2bddca658976
+ATSRC_PACKAGE_REV=2052837a3e0b
 ATSRC_PACKAGE_LICENSE="GPL 2.0"
 ATSRC_PACKAGE_DOCLINK="http://valgrind.org/docs/"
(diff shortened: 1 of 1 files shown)


configs/next/packages/valgrind/sources
```diff
@@ -51,6 +51,11 @@ atsrc_get_patches ()
 	at_get_patch \
 		'https://raw.githubusercontent.com/powertechpreview/powertechpreview/master/Valgrind%20iTrace%20Patches/3.14/update_itrace_to_use_epoch.v2.diff' \
 		41fc03e4b074698f1fff9e7a53fed63a || return ${?}
+
+	# Update itrace to Valgrind 3.15.
+	at_get_patch \
+		'https://raw.githubusercontent.com/powertechpreview/powertechpreview/master/Valgrind%20iTrace%20Patches/3.15/Fix-itrace-initialization-of-guest_chase-variable.patch' \
```
rzinsly: I suggest to use https://raw.githubusercontent.com/powertechpreview/powertechpreview/0496d9de5cf3d441787d413367c5ec514a64b2ae/Valgrind%20iTrace%20Patches/3.15/Fix-itrace-initialization-of-guest_chase-variable.patch instead, it's better to use a fix commit ID in case that file is updated and breaks the build.

er-1: Good point. Thx.

PR #1299, opened 2020-02-14
Force GCC stage 2 bootstrapping to use AT's libc

### LINKED ISSUE #1298: AT fails to build on RHEL 8
I've now tried to build on two RHEL 8.1 systems, and they both fail identically:
```
...
Starting the build of gcc_2...
make: *** [Makefile:1342: /home/pc/advance-toolchain/at14.0-0-alpha.redhat-8_ppc64le_ppc64le/receipts/gcc_2.b.rcpt] Error 1
```
`at14.0-0-alpha.redhat-8_ppc64le_ppc64le/logs/_gcc_2-3_standard_buildf-06_make.log`:
```
configure: error: in `/home/pc/advance-toolchain/at14.0-0-alpha.redhat-8_ppc64le_ppc64le/builds/gcc_2/libiberty':
configure: error: cannot run C compiled programs.
If you meant to cross compile, use `--host'.
See `config.log' for more details
checking for string.h... /opt/at-next-14.0-0-alpha/bin/gcc -E
yes
checking for sys/stat.h... make[3]: *** [Makefile:10923: configure-stage1-libiberty] Error 1
```
`at14.0-0-alpha.redhat-8_ppc64le_ppc64le/builds/gcc_2/libiberty/config.log`:
```
configure:3495: checking whether we are cross compiling
configure:3503: /opt/at-next-14.0-0-alpha/bin/gcc -o conftest -g  -Wl,--dynamic-linker=/opt/at-next-14.0-0-alpha/lib64/ld64.so.2  conftest.c  >&5
configure:3507: $? = 0
configure:3514: ./conftest
/home/pc/advance-toolchain/at14.0-0-alpha.redhat-8_ppc64le_ppc64le/sources/gcc/libiberty/configure: line 3516: 103800 Segmentation fault      ./conftest$ac_cv_exeext
configure:3518: $? = 139
configure:3525: error: in `/home/pc/advance-toolchain/at14.0-0-alpha.redhat-8_ppc64le_ppc64le/builds/gcc_2/libiberty':
configure:3527: error: cannot run C compiled programs.
```
And indeed, the newly built linker does not produce working executables:
```
$ echo 'int main(){}' > conftest.c
$ /opt/at-next-14.0-0-alpha/bin/gcc -o conftest -g  -Wl,--dynamic-linker=/opt/at-next-14.0-0-alpha/lib64/ld64.so.2  conftest.c
$ ./conftest
Segmentation fault (core dumped)
```
It seems to be caused by using the newly built loader, the newly built exectable, and the _system_ libc:
```
$ ldd ./conftest
        linux-vdso64.so.1 (0x00007fff85e00000)
        libc.so.6 => /lib64/power9/libc.so.6 (0x00007fff85bd0000)
        /opt/at-next-14.0-0-alpha/lib64/ld64.so.2 => /lib64/ld64.so.2 (0x00007fff85e20000)
```
Manually adding an `rpath` which finds the newly built libc instead produces a working executable:
```
$ /opt/at-next-14.0-0-alpha/bin/gcc -o conftest -g  -Wl,--dynamic-linker=/opt/at-next-14.0-0-alpha/lib64/ld64.so.2  conftest.c -Wl,-rpath=/opt/at-next-14.0-0-alpha/lib64
$ ldd ./conftest
        linux-vdso64.so.1 (0x00007fffbdf90000)
        libc.so.6 => /opt/at-next-14.0-0-alpha/lib64/libc.so.6 (0x00007fffbdd70000)
        /opt/at-next-14.0-0-alpha/lib64/ld64.so.2 => /lib64/ld64.so.2 (0x00007fffbdfb0000)
$ ./conftest
$ 
```

mscastanho:
**EDIT**: this PR is based on the solution suggested by @ThinkOpenly [here](https://github.com/advancetoolchain/advance-toolchain/issues/1298#issuecomment-584386418).

@ThinkOpenly, this is intended to fix issue #1298.

```
The build of GCC stage 2 is configured to use --with-advance-toolchain
to guarantee that the linker and headers from AT are used instead of the
system's, which includes setting the -rpath linker flag.

However, this setting does not apply to the tools built during GCC's
bootstrap process. The dynamic linker is currently using the system's
libc.so and AT's ld.so, which can cause problems.

This commit sets the correct -rpath to be used by the linker on stage 1
of the bootstrap to fix this issue.
```

Extra considerations:
- I actually tested setting the rpath to `${at_dest}/lib64`, only after I realized that using `${libdir}` is probably better to handle `lib` / `lib64` properly. This PR's automated test will check if this also works =)
- Should I restrict this change only to AT 13 and AT next? Currently it applies all the way back to 11.0
- I'm not completely sure if we also need to use `--with-boot-ldflags` to apply the `-rpath` flag to the other bootstrap stages. Not sure also if this change should apply to AT's GCC stages 3, 4 and opt as well. My test didn't need any of those to pass.

### DIFF
--- a/configs/11.0/packages/gcc/stage_2
+++ b/configs/11.0/packages/gcc/stage_2
@@ -121,7 +121,7 @@ atcfg_configure() {
 			--with-mpc-lib=${libdir} \
 			--with-as="${at_dest}/bin/as" \
 			--with-ld="${at_dest}/bin/ld" \
-			--with-stage1-ldflags="-Wl,--dynamic-linker=${ldso}"
+			--with-stage1-ldflags="-Wl,--dynamic-linker=${ldso},-rpath=${libdir}"
 	else
 		# Configure command for cross builds
 		CC="${system_cc}" \


mscastanho:
As noted in #1298, this PR does not fix the issue on latest AT version and will likely need changes. Since @ThinkOpenly is taking a look at this issue again, I'm closing this.

PR #1302, opened 2020-02-15
Update gcc on AT next

Cherbot:
Bump to revision d6ee2e7c5a6b


er-1:
It failed to apply the patch `0001-Skip-machine_name-fix-for-powerpc-on-fixincludes-ste.patch` on gcc source code.

mscastanho:
@er-1 the PR mentioned above should fix this problem. I left AT 13.0 untouched as the patch was merged upstream only to the master branch.

er-1:
This was fixed by #1309

PR #1308, opened 2020-02-17
Fix split of binutils on AT 13.0 from AT next

er-1:
Set the correct commit id from inside the 2.34 branch.


mscastanho:
I can't seem to find the 2.34 branch. `git branch -a` shows just up to `binutils-2_33-branch`

er-1:
Snippet of the output of `git ls-remote git://sourceware.org/git/binutils-gdb.git`:
```
refs/heads/binutils-2_30-branch
refs/heads/binutils-2_31-branch
refs/heads/binutils-2_32-branch
refs/heads/binutils-2_33-branch
refs/heads/binutils-2_34-branch
```

mscastanho:
Thanks, now I got it.
 
Not sure if this makes any difference, but commit `d7f734bc7e9e` sets the version to `2.34`. Does AT need the extra `.0`?

Besides this, LGTM.

er-1:
> Not sure if this makes any difference, but commit `d7f734bc7e9e` sets the version to `2.34`. Does AT need the extra `.0`?

The version is checked against the content of `bfd/version.m4`: `m4_define([BFD_VERSION], [2.34.0])`
It's a way to keep track of small changes in a given version.

mscastanho:
Thanks for the explanation =)

PR #1309, opened 2020-02-17
Remove GCC fixincludes patch on AT next

mscastanho:
The out-of-tree patch to GCC to skip the machine_name fixincludes
fix is not needed anymore on AT next, as it has been merged upstream
on master as d876c8f9cfbf2b670bca9bc15bb2f472de5da57.

This should fix the build problems seen with #1302.


rzinsly:
You mention that the fix have been merged with d876c8f9cfbf2b670bca9bc15bb2f472de5da57 but this sources still uses 561d91dcc911.
You can update the revision here so this patch doesn' t break the build until next update by Cherbot.

mscastanho:
@rzinsly Nice catch, thanks. This PR was intended to be merged with #1302, as it already uses a more up-do-date revision that includes the patch I mentioned.

But it's probably better to just pack everything on one PR instead. I just updated this PR with a cherry-pick of the commit from #1302 to properly update GCC.

er-1:
I agree. This will close #1302 

Close #1302

Issue #1383, opened 2020-04-07
Distribute libnxz on AT 14.0


tuliom:
[libnxz](https://github.com/libnxz/power-gzip) is a library that is API and ABI-compatible with zlib and provides access to the NX GZIP processor available on POWER9 processors.
It's dual-licensed Apache 2.0 and GPLv2.

We'd need version >= 0.60.
I suggest to package it on its own RPM/DEB as it may not be used by all users.

tuliom:
Notice this package is not required on AT <= 13.0.

tuliom:
Reminder: new packages are usually followed by a new FVTR testcase  ;-)

er-1:
@lucmaga The issue #1450 is related with this.

er-1:
#1561 fixes this.

PR #1395, opened 2020-04-15
Add libnxz to AT next

lucmaga:


er-1:
@lucmaga whata about new FVTR testcase? Most of the time, new package comes with new FVTR testcase ;)

lucmaga:
Thanks @er-1. I just realize I forgot to install the include files as well. Sorry about that.

configs/next/packages/libnxz/stage_1
```diff
@@ -0,0 +1,106 @@
...
+}
+
+atcfg_pre_install() {
+	mkdir -p ${install_transfer}/lib64/
+}
+
+atcfg_install() {
+	# libnxz does not have a make install
+	PATH=${at_dest}/bin:${PATH} \
+		install -m 755 ${ATSRC_PACKAGE_WORK}/libnxz.so ${install_transfer}/lib64/libnxz.so
```
tuliom, line 104: Likewise for `configure_libdir`.

This is also missing `libnxz.a` and probably a header.  Need to confirm.

Notice that newer versions (still in develop branch) will require changes to this line.

configs/next/packages/libnxz/stage_1
```diff
@@ -0,0 +1,106 @@
...
+		PATH=${at_dest}/bin:${PATH} ${SUB_MAKE} check -C ${ATSRC_PACKAGE_WORK}
+	fi
+}
+
+atcfg_pre_install() {
+	mkdir -p ${install_transfer}/lib64/
+}
+
+atcfg_install() {
+	# libnxz does not have a make install
```
tuliom, line 102: Reported [here](https://github.com/libnxz/power-gzip/issues/11).

configs/next/packages/libnxz/stage_1
```diff
@@ -0,0 +1,106 @@
...
+	# not, e.g.:
+	#   configure --build=powerpc64-linux-gnu --host=powerpc-linux-gnu
+	#
+	# Therefore, on native builds, we force configure_build to have the
+	# same value as configure_host.
+	configure_host=$(find_build_target ${AT_BIT_SIZE})
+	if [[ "${cross_build}" == 'no' ]]; then
+		configure_build=${configure_host}
+	else
+		configure_build=${host}
```
tuliom, line 54: Where is `configure_build` being used?

tuliom:
By the way, this kind of patch deserves a good commit message describing how the license issues are being dealt with in the release notes and source code tarball, which package will distribute libnxz and how it was tested.

tuliom:
Confirmed: the `Legal notices` section from the release notes needs an update.
libnxz should be listed in the GPL section and in the following section with a copy of the Apache 2.0 license.

configs/next/packages/libnxz/sources
```diff
@@ -0,0 +1,42 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# POWER NX zlib compliant library source package and build info
+# ==============================================================
+#
+
+ATSRC_PACKAGE_NAME="POWER NX zlib compliant library"
```
tuliom, line 22: What about `NX GZIP library` instead?
I don't think "zlib" and "compliant" should be part of the name.

tuliom:
I believe that scripts/utilities/check_tarball.sh:25 needs to be updated. `src_list` must include all *GPL packages.

configs/next/packages/libnxz/sources
lucmaga: I'm ok with changing it. I was using the same name from the projects github page.

lucmaga:
I'm still going to install on a server for testing.

configs/next/packages/libnxz/stage_1
```diff
@@ -0,0 +1,72 @@
...
+atcfg_pre_install() {
+	mkdir -p ${install_transfer}/${configure_libdir}
+	mkdir -p ${install_transfer}/include/libnxz
+}
+
+atcfg_install() {
+	# libnxz does not have a make install
+	install -m 755 ${ATSRC_PACKAGE_WORK}/libnxz.so ${install_transfer}/${configure_libdir}/libnxz.so
+	install -m 644 ${ATSRC_PACKAGE_WORK}/libnxz.a ${install_transfer}/${configure_libdir}/libnxz.a
+	install -m 644 ${ATSRC_PACKAGE_WORK}/inc_nx/* ${install_transfer}/include/libnxz/
```
tuliom, line 69: I don't think this line is correct.
@rzinsly Any idea which headers should be installed?

configs/next/packages/libnxz/stage_1
```diff
@@ -0,0 +1,72 @@
...
+}
+
+atcfg_make() {
+	# In gcc10 the default is now -fno-common. In order to not have multiple
+	# definitions -fcommon is needed.
+	configure_cc="${at_dest}/bin/${target64:-${target}}-gcc"
+	PATH="${at_dest}/bin:${PATH}" \
+	${SUB_MAKE} -C ${ATSRC_PACKAGE_WORK} \
+	CC="${configure_cc} -m${AT_BIT_SIZE}" \
+	FLG="-std=gnu11 -fcommon" ZLIB=""
```
tuliom, line 50: Why is this removing `ZLIB = -DZLIB_API` ?

configs/next/packages/libnxz/stage_1
```diff
@@ -0,0 +1,72 @@
...
+atcfg_make_check() {
+	# Package testing is not done on a cross build.
+	if [[ "${cross_build}" == 'no' ]]; then
+		PATH=${at_dest}/bin:${PATH} ${SUB_MAKE} check -C ${ATSRC_PACKAGE_WORK}
+	fi
+}
+
+atcfg_pre_install() {
+	mkdir -p ${install_transfer}/${configure_libdir}
+	mkdir -p ${install_transfer}/include/libnxz
```
tuliom, line 62: Is this installing the headers in the right sysroot path for a cross compiler build?

configs/next/packages/libnxz/stage_1
rzinsly: nx-zlib.h, note that it will be under lib/ in next releases
Also the ones in inc_nx if you want to be able to run the codes at samples/, note that some use internal symbols. At least if we want the internal debug log routines inc_nx/nx_dbg.h.

configs/next/packages/libnxz/stage_1
rzinsly: ZLIB = -DZLIB_API will substitute the zlib API by libnxz's, removing that and you will have libnxz standalone and can have both libraries.

rzinsly:
IMHO it's ok to put libnxz in the toolchain_extra bundle but it would be better in its own package like sugested in https://github.com/advancetoolchain/advance-toolchain/issues/1383

configs/next/packages/libnxz/stage_1
lucmaga: I think everything is covered.

configs/next/packages/libnxz/stage_1
lucmaga: Is there any difference?
For what I understand "${install_transfer}" expands to  "$DESTDIR/at-${version}/"
Anyway I'm still looking for the cross compile problems.

configs/next/packages/libnxz/stage_1
tuliom: @rzinsly Why can't we have both libraries providing the same symbols?

configs/next/packages/libnxz/stage_1
tuliom: @lucmaga my question is specific to cross compilers. If you leave a header out of the sysroot, the cross compiler won't find it.

configs/next/packages/libnxz/stage_1
tuliom: > Also the ones in inc_nx if you want to be able to run the codes at samples/

Sounds like these headers should not be installed.

lucmaga:
> IMHO it's ok to put libnxz in the toolchain_extra bundle but it would be better in its own package like sugested in #1383

Any name suggestions ? toolchain_extra_libnxz or toolchain_libnxz or just libnxz ?

configs/next/packages/libnxz/stage_1
rzinsly: @tuliom: If we distribute both libraries I guess we should be able to use both, right?
I recognize that it would not be a common usage anyway.

rzinsly:
> > IMHO it's ok to put libnxz in the toolchain_extra bundle but it would be better in its own package like sugested in #1383
> 
> Any name suggestions ? toolchain_extra_libnxz or toolchain_libnxz or just libnxz ?

Just libnxz seems fine, using toolchain in the name may cause confusion with the others as it won't be part of the runtime package.

configs/next/packages/libnxz/stage_1
tuliom: @rzinsly Yes.  We won't stop distributing zlib.  We need to have both.
AFAIU, most users will not be interested on libnxz in the beginning, so we should not install it automatically.
But when they become interested, they have to be able to easily replace zlib with libnxz without having to recompile their software.  Assuming it has already been built with libnxz.

rzinsly: > they have to be able to easily replace zlib with libnxz without having to recompile their software. Assuming it has already been built with libnxz.

I didn't understand this bit, to build with libnxz they would have to recompile the software, right? They can just use LD_PRELOAD but that is not the easiest way IMO.
I guess you mean that the user will not want to change the code to start using the nx_ symbols? It makes sense, I'm ok with keeping -DZLIB_API if that is the case. But I don't see a problem on having it providing only the nx_ symbols for a full standalone library, also it could be used in the unlikely scenario that someone wants to use the NX accelerator for some things and zlib for others in the same application.
I'm ok with whatever the team decides it's the behaviour the library should have on AT.

tuliom: @rzinsly I'm sorry. I meant to say: "Assuming it has already been built with AT".
If they have built their software with AT, they can use `LD_PRELOAD` to replace zlib with libnxz.
Which way do you think is easiest for customers?

> I guess you mean that the user will not want to change the code to start using the nx_ symbols?

Exactly.

> in the unlikely scenario that someone wants to use the NX accelerator for some things and zlib for others in the same application.

In order to do this, this user would have to change their code, regardless of which symbols are exported in libnxz.
IMHO, it's technically possible to do this even when libnxz is compiled with `-DZLIB_API` provided that upstream libnxz is prepared for this. If it isn't we can always submit bug reports upstream.

configs/next/packages/libnxz/stage_1
```diff
@@ -0,0 +1,77 @@
...
+	fi
+	mkdir -p ${install_dir}/${configure_libdir}
+	mkdir -p ${install_dir}/include/libnxz
+}
+
+atcfg_install() {
+	# libnxz does not have a make install
+	install -m 755 ${ATSRC_PACKAGE_WORK}/libnxz.so ${install_dir}/${configure_libdir}/libnxz.so
+	install -m 644 ${ATSRC_PACKAGE_WORK}/libnxz.a ${install_dir}/${configure_libdir}/libnxz.a
+	install -m 644 ${ATSRC_PACKAGE_WORK}/inc_nx/* ${install_dir}/include/libnxz/
```
tuliom, line 74: AFAIU, this line should be removed.
More details here: https://github.com/advancetoolchain/advance-toolchain/pull/1395#discussion_r412152299

configs/next/packages/libnxz/stage_1
rzinsly: > If they have built their software with AT, they can use `LD_PRELOAD` to replace zlib with libnxz.
> Which way do you think is easiest for customers?

I didn't mean to say that use LD_PRELOAD is difficult, I just don't think that exporting or keep setting it is not easier than recompile the program once even for big applications.

configs/next/packages/libnxz/stage_1
lucmaga: It's not clear for me from that discussion this line should be removed. From what I understood
this include files might be necessary in some cases.
I never used libnxz so I'm relying only on the above discussion.

configs/next/packages/libnxz/stage_1
lucmaga: Indeed it was wrong. Fixed now.

lucmaga:
> I'm missing changes to the RPM specfile.

Done

> The name of the package must have the AT major version.
> Without this, users won't be able to install libnxz from AT 14.0 and libnxz from AT 15.0 on the same system.
> Each one has different runtime requirements.

Done

configs/next/packages/libnxz/stage_1
lucmaga: @tuliom, @rzinsly The codes at sample/ aren't examples usage? 
If so, either the includes are essential.

configs/next/packages/libnxz/stage_1
lucmaga: So I will keep the -DZLIB_API.

configs/next/packages/libnxz/stage_1
tuliom: @lucmaga They are indeed examples. IMHO, they are **not** state of the art examples, though.
They're using non-exported functions and internal structures that should not be used in examples.
IMHO, real user code should not be written that way.
With that said, I do believe I could be missing some functions or structures that should indeed be made public.  However, it's easier to not export them now (i.e. without installing headers) and start export them after we confirm the mistake.  It's more complicated to stop exporting an internal structure or function after users started depending on them.

lucmaga: > However, it's easier to not export them now (i.e. without installing headers) and start export them after we confirm the mistake. It's more complicated to stop exporting an internal structure or function after users started depending on them.

Agreed.

lucmaga:
@tuliom, @rzinsly 
I needed to add inc_nx/nxu.h and inc_nx/nx_dbg.h they are dependencies for nx_zlib.h

tuliom:
@lucmaga I've just remembered yet another file change missing: https://github.com/advancetoolchain/advance-toolchain/blob/master/utils/install_tools.sh
Sorry.

@er-1 any other file changes needed related to automated tasks?

er-1:
> @er-1 any other file changes needed related to automated tasks?
I see nothing to add.
I think a change will be needed there https://github.ibm.com/advancetoolchain/at-utils ; after the merge of this PR.

lucmaga:
@tuliom the failed build was actually a success looking on the CI interface.

tuliom:
> @tuliom the failed build was actually a success looking on the CI interface.

I reported the build failure as #1438 and requested a new AT 13.0 RHEL8 ppc64le build.

configs/next/packages/libnxz/stage_1
lucmaga: It was removed and replaced for only includes needed for nx_zlib.h

lucmaga:
> The last paragraph from this comment has not been implemented:
> [#1395 (review)](https://github.com/advancetoolchain/advance-toolchain/pull/1395#pullrequestreview-399492880)

About this dummy package. I can't see a real usage for it. @tuliom if you really think it is useful I
would suggest that it should be done in another PR. So we add a dummy package for libnxz and for the
other packages as well at once.

> This comment is still an issue: [#1395 (comment)](https://github.com/advancetoolchain/advance-toolchain/pull/1395#discussion_r414139049)

It was already solved for this PR. Although there is still a discussion if we should wait for libnxz to fix it's include files. See [#1395 (comment)](https://github.com/advancetoolchain/advance-toolchain/pull/1395#issuecomment-620586420)

lucmaga:
> LGTM overall, I have a cosmetic suggestion:
> Change the package description in the native specfile, it looks like the perf package description but I don't think it has a good one, I believe it's better to use the same line you used in the cross package: "This package provides the NX GZIP Library."
>

Agreed. Done.

> And when you need to replace symlinks in the future I suggest to create a commit that only replace them before a commit that actually does the changes (like [ed324c6](https://github.com/advancetoolchain/advance-toolchain/commit/ed324c6e42242fc24af47b575c7b3c69056d28d5)), this make it easier to review.

Thanks for the advice.

tuliom:
> About this dummy package. I can't see a real usage for it.

@lucmaga , allow me to repeat myself so that you understand why it's needed:

>> ... it's a good idea to provide a dummy package, without a version in the name, so that it's easier for software depending on AT, to require AT packages, e.g. the SDK.

> @tuliom if you really think it is useful I
> would suggest that it should be done in another PR. So we add a dummy package for libnxz and for the
> other packages as well at once.

This is not a new thing for DEB. libnxz is the only package that does not provide the dummy package, e.g.:
https://github.com/advancetoolchain/advance-toolchain/pull/1395/files#diff-3a9df9b0eef7ed67c1b78997b88eca6dR141

tuliom:
@lucmaga I've just realized the debug packages are also missing for libnxz on DEB.

lucmaga:
Done, the dummy package and the debug packages.

PR #1402, opened 2020-04-18
Update valgrind on AT next

Cherbot:
Bump to revision 6f84421546ea


er-1:
Valgrind FVTR failed:
`valgrind: power9:  library /home/tcbot/bot-slave/at_opt/at-next-14.0-0-alpha/lib/valgrind/vgpreload_exp-sgcheck-ppc64le-linux.so  is missing`

tuliom:
That tool has been removed with commit:

```
commit 40187fcd61ee877f78701c46a74ac1dadbe65b3d
Author: Julian Seward <jseward@acm.org>
Date:   Fri Apr 17 19:23:26 2020 +0200

    Remove the exp-sgcheck tool.
    
    It only ever worked on x86 and amd64, and even on those it had a high false
    positive rate and was slow.  Everything it does, ASan can do faster, better,
    and on more architectures.  So there's no reason to keep this tool any more.
```

Looks like we don't even need to test it on older Valgrind versions.

PR #1415, opened 2020-04-25
Update gcc on AT 12.0

Cherbot:
Bump to revision e25f27eea473


mscastanho:
The `ck_binaries` check on the testsuite failed on all builds.

PR #1414, opened 2020-04-25
Update gcc on AT 13.0

Cherbot:
Bump to revision 200e83ac36b7

PR #1420, opened 2020-04-28
Update gcc on AT 13.0

er-1:
Bump to version 9.3.1
Bump to revision 200e83ac36b7

Close #1414 


mscastanho:
I believe this is good to go. But perhaps we should squash the two commits into one so we don't introduce a broken commit into the project's history.

er-1:
> 
> 
> I believe this is good to go. But perhaps we should squash the two commits into one so we don't introduce a broken commit into the project's history.

Done.

PR #1435, opened 2020-05-09
Update glibc on AT next

Cherbot:
Bump to revision e1b871e25f3a

PR #1443, opened 2020-05-16
Update glibc on AT next

Cherbot:
Bump to revision 61d64408a1f4


mscastanho:
As I pointed out in #1435, I believe this is being caused by `06febd8c67` on glibc.
I just added a comment to the original bugzilla issue related to that commit ([link](https://sourceware.org/bugzilla/show_bug.cgi?id=25219#c3)).

er-1:
Closing. We'll see if this is fixed next Saturday.

Issue #1450, opened 2020-05-20
AT next can't be installed


er-1:
When installing AT next (from packages), the following issue occurs:
```
Selecting previously unselected package advance-toolchain-at14.0-alpha-devel-dbg.
(Reading database ... 122317 files and directories currently installed.)
Preparing to unpack advance-toolchain-at14.0-alpha-devel-dbg_14.0-0-alpha_ppc64el.deb ...
Unpacking advance-toolchain-at14.0-alpha-devel-dbg (14.0-0-alpha) ...
Selecting previously unselected package advance-toolchain-at14.0-alpha-libnxz-dbg.
dpkg: regarding advance-toolchain-at14.0-alpha-libnxz-dbg_14.0-0-alpha_ppc64el.deb containing advance-toolchain-at14.0-alpha-libnxz-dbg, pre-dependency problem:
 advance-toolchain-at14.0-alpha-libnxz-dbg pre-depends on advance-toolchain-at14.0-alpha-libnxz (= 14.0-0-alpha)
  advance-toolchain-at14.0-alpha-libnxz is not installed.
dpkg: error processing archive advance-toolchain-at14.0-alpha-libnxz-dbg_14.0-0-alpha_ppc64el.deb (--install):
 pre-dependency problem - not installing advance-toolchain-at14.0-alpha-libnxz-dbg
Selecting previously unselected package advance-toolchain-at14.0-alpha-mcore-libs-dbg.
Preparing to unpack advance-toolchain-at14.0-alpha-mcore-libs-dbg_14.0-0-alpha_ppc64el.deb ...
Unpacking advance-toolchain-at14.0-alpha-mcore-libs-dbg (14.0-0-alpha) ...
Selecting previously unselected package advance-toolchain-at14.0-alpha-perf-dbg.
Preparing to unpack advance-toolchain-at14.0-alpha-perf-dbg_14.0-0-alpha_ppc64el.deb ...
Unpacking advance-toolchain-at14.0-alpha-perf-dbg (14.0-0-alpha) ...
Selecting previously unselected package advance-toolchain-at14.0-alpha-runtime-dbg.
Preparing to unpack advance-toolchain-at14.0-alpha-runtime-dbg_14.0-0-alpha_ppc64el.deb ...
Unpacking advance-toolchain-at14.0-alpha-runtime-dbg (14.0-0-alpha) ...
Setting up advance-toolchain-at14.0-alpha-devel-dbg (14.0-0-alpha) ...
[...]

PR #1451, opened 2020-05-20
Fix install_tools.sh debian package order

lucmaga:
Fix issue #1450.

### DIFF
--- a/utils/install_tools.sh
+++ b/utils/install_tools.sh
@@ -34,12 +34,12 @@ install_native ()
 		sudo dpkg -i advance-toolchain-at*devel_* \
 			advance-toolchain-at*mcore-libs_* \
 			advance-toolchain-at*perf_* || return ${?}
+		if [ -e advance-toolchain-at*libnxz_* ]; then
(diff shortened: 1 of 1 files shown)


utils/install_tools.sh
```diff
@@ -34,13 +34,13 @@ install_native ()
 		sudo dpkg -i advance-toolchain-at*devel_* \
 			advance-toolchain-at*mcore-libs_* \
 			advance-toolchain-at*perf_* || return ${?}
+		if [ -e advance-toolchain-*libnxz* ]; then
```
er-1, line 37: This matches more than one file, so the test fails, and the libnxz package is not installed. That's why when installing the dbg packages it fails about libnxz-dbg package.

So I'd put this : `-e advance-toolchain-at*libnxz_*`

utils/install_tools.sh
```diff
@@ -34,13 +34,13 @@ install_native ()
...
+			sudo dpkg -i advance-toolchain-*libnxz_* || return ${?}
```
er-1, line 38: I'd put this: `advance-toolchain-at*libnxz_*`

PR #1454, opened 2020-05-23
Update gcc on AT next

Cherbot:
Bump to revision b2b8eb6202b8


mscastanho:
The problem was caused by an internal compiler error (ICE) when compiling the tuned glibc. The same error has already been reported upstream on GCC, and is apparently already fixed ([Link](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95273)). 

So I'll close this now and let Cherbot bump the version over the weekend.

Issue #1472, opened 2020-06-05
Diverge AT 14.0 from next


er-1:
GCC 10.2 is scheduled for July 2020.

The GCC 10 branch is about to be created and then will start to diverge from trunk.
It's time to also diverge the AT 14.0 config set from AT next.

er-1:
This depends on:
- #1551
- #1552
- #1553
- #1554
- #1555
- #1556
- #1557

mscastanho:
@er-1 All the items this issue depended on have been closed. Can we close this one too?

er-1:
@mscastanho I'd like to wait for the GA release of AT 14.0-0 before closing this.

er-1:
AT 14.0-0 was released 3 days ago.

PR #1488, opened 2020-06-15
Prepare AT 14.0

er-1:
Clone AT next into AT 14.0 and update version number in configuration files.
From there, each package can be set up to diverge from AT next.

This is related with #1472


er-1:
> Not completely related to this patch, but I wonder if the support for Ubuntu 20.04 is good enough so that we can stop building for 18.04.

@tuliom Do you think we could provide AT 14.0 on both 18.04 and 20.04? So we'll still have 18.04 in case we have to "drop" 20.04.

tuliom:
> Do you think we could provide AT 14.0 on both 18.04 and 20.04?

@er-1 Sure!

PR #1499, opened 2020-06-25
Update packages for AT 14.0

er-1:
This is related with #1472


configs/14.0/packages/glibc/sources
```diff
@@ -1,53 +0,0 @@
-#!/usr/bin/env bash
```
mscastanho, line 1: It caught my attention to see 14 pointing to a future version (next). But it's explained in the commit message. Makes sense. Thanks!

configs/14.0/packages/binutils/sources
```diff
@@ -1,55 +0,0 @@
-#!/usr/bin/env bash
```
mscastanho, line 1: So there hasn't been a new binutils release since 13.0 was out?

configs/14.0/packages/gcc/sources
```diff
@@ -54,8 +56,8 @@ atsrc_get_patches ()
 
 	# Avoid a misconfiguration of tuned libraries as seen in issue #204.
 	at_get_patch \
-		'https://raw.githubusercontent.com/powertechpreview/powertechpreview/704c09577104f48acca9eb267e5c5afda183eb71/GCC%20PowerPC%20Backport/9/0001-libssp-Ignore-vsnprintf-test-when-ssp_have_usable_vs.patch' \
-		f62370874f2717c6e530d27083434fe8 || return ${?}
+		'https://raw.githubusercontent.com/powertechpreview/powertechpreview/a51115f619a5210d868819d41ba5f021327ba83/GCC%20PowerPC%20Backport/9/0001-libssp-Ignore-vsnprintf-test-when-ssp_have_usable_vs.patch' \
+		8ad129baf6f4a6f61572d5c36c57536b || return ${?}
```
mscastanho, line 60: Use the previous libssp patch. OK.

configs/14.0/packages/gcc/sources
```diff
@@ -22,18 +22,20 @@
 ATSRC_PACKAGE_NAME="GCC (GNU Compiler Collection)"
 ATSRC_PACKAGE_SUBPACKAGE_NAME_0="GNU Standard C++ Library v3 (Libstdc++-v3)"
 ATSRC_PACKAGE_SUBPACKAGE_NAME_1="GNU Libgomp"
-ATSRC_PACKAGE_VER=11.0.0
-ATSRC_PACKAGE_REV=2e6a6644db0e
+ATSRC_PACKAGE_VER=10.1.1
+ATSRC_PACKAGE_REV=fa34f36ef5e
 ATSRC_PACKAGE_LICENSE="GPL 3.0"
-ATSRC_PACKAGE_DOCLINK="https://gcc.gnu.org/onlinedocs/gcc/"
+ATSRC_PACKAGE_DOCLINK="https://gcc.gnu.org/onlinedocs/gcc-${ATSRC_PACKAGE_VER%.*}.0/gcc/"
```
mscastanho, line 28: URL working as expected. OK.

mscastanho:
(_For some weird reason GitHub is not allowing me to do an inline comment, it keeps saying `Line must be part of the diff`. Oh well..._)

I noticed that the patch below is not being applied by `atsrc_apply_patches` later. The same for the 14.0 version. Can we stop fetching it or should it actually be applied?
https://github.com/advancetoolchain/advance-toolchain/blob/2c00611d35bf80c9fc79b2d96457f13c982bd6f7/configs/next/packages/gcc/sources#L50-L53

configs/14.0/packages/binutils/sources
er-1: > So there hasn't been a new binutils release since 13.0 was out?

Right.
That said, binutils 2.35 could be released in July. I'll update my changes.

er-1:
> I noticed that the patch below is not being applied by `atsrc_apply_patches` later. The same for the 14.0 version. Can we stop fetching it or should it actually be applied?

It's applied, but only  on stage 4.
https://github.com/advancetoolchain/advance-toolchain/blob/ecc0e78ae68074ec3e6be875750cc0e7abb6f4d4/configs/13.0/packages/gcc/stage_4#L139-L157

PR #1508, opened 2020-06-29
Update more packages for AT 14.0

er-1:
Related with #1472


mscastanho:
https://github.com/advancetoolchain/advance-toolchain/blob/d63b0be85e1ac6d775672e4c47a98666a1a1c292/configs/next/packages/libsphde/sources#L26
This URL is not working. We should replace it by https://sphde.github.io/

mscastanho:
The changes LGTM. I noticed that AT 14.0 and next are still following Python 3.7 branch, while 3.8 is already out. Should we update that as well?

er-1:
> The changes LGTM. I noticed that AT 14.0 and next are still following Python 3.7 branch, while 3.8 is already out. Should we update that as well?

I think it deserves its own PR. So it'll come after this one.

mscastanho:
libnxz 0.61 has just been released ([link](https://github.com/libnxz/power-gzip/releases/tag/v0.61)), so commit https://github.com/advancetoolchain/advance-toolchain/pull/1508/commits/5fc4cb9619a917d17e24f8eb9185d32dd8266e91 can be updated

er-1:
> libnxz 0.61 has just been released , so commit [5fc4cb9](https://github.com/advancetoolchain/advance-toolchain/commit/5fc4cb9619a917d17e24f8eb9185d32dd8266e91) can be updated

Good news.
I propose to do that in another PR.

PR #1528, opened 2020-07-04
Update expat on AT 13.0

Cherbot:
Bump to revision 253686453276


er-1:
This would bump the version of expat out of the AT 13.0 scope.

PR #1529, opened 2020-07-04
Update libhugetlbfs on AT 13.0

Cherbot:
Bump to revision e6499ff92b4a


er-1:
This would bump the version of libhugetlbfs out of the AT 13.0 scope.

PR #1538, opened 2020-07-04
Update valgrind on AT 12.0

Cherbot:
Bump to revision efbc655366e3


er-1:
This would bump the version of valgrind out of the AT 12.0 scope.

PR #1541, opened 2020-07-07
Remove flag to ignore string overflow on glibc

mscastanho:
Starting with version 10, GCC started yielding false-positives related
to string overflows on iconv-related code on glibc. Because of that, we
added the -Wno-error=stringop-overflow flag to glibc configuration
to suppress the warning and let the builds complete.

This issue has just been fixed upstream by commit 08538f360f, so now we
can remove that flag from all glibc stages.


mscastanho:
All tests passed but I had to force push to resolve conflicts with master branch.

PR #1546, opened 2020-07-09
Fix gdbserver build on AT 14.0

er-1:
Previous commit #2781f58 fixed the build of gdbserver on AT next.
AT 14.0 needs this fix too.


er-1:
Not a fix at all.

PR #1549, opened 2020-07-17
Make debuginfo usable by `perf`

### LINKED ISSUE #965: debuginfo file locations are not generally useful
The Advance Toolchain debuginfo packages place their files in `/opt/at{version}/lib/debug/opt/at{version}/lib64`. To be generally useful, for example by `gdb` or `perf`, the following algorithm should be observed (from https://sourceware.org/gdb/onlinedocs/gdb/Separate-Debug-Files.html).

For a library `/opt/atversion/lib64/libA.so`, with "debug link" `libA.so.debug` and "build-id" `abcdef`, its corresponding debuginfo file will be looked for at the following locations, in order:
1. `/usr/lib/debug/.build-id/ab/cdef/.debug`
1. `/opt/atversion/lib64/libA.so.debug`
1. `/opt/atversion/lib64/.debug/libA.so.debug`
1. `/usr/lib/debug/opt/atversion/lib64/libA.so.debug`

The current location of the Advance Toolchain debuginfo files will not match any of those locations.

If there is a strong desire to avoid putting files in `/usr`, then the following options are reasonable locations that will work in general:
1. `/opt/atversion/lib64/libA.so.debug`
1. `/opt/atversion/lib64/.debug/libA.so.debug`

_Note that the `gdb` that is included with Advance Toolchain works because in addition to `/usr/lib/debug`, it is apparently modified to include an additional "global debug directory" of `/opt/at{version}/lib/debug`._

All that being said, an `strace` of `perf` shows the only attempt to find a debuginfo file in `/opt` was for `/opt/at{version}/lib64/.debug/libA.so`.  (Note there is no `.debug` appended.)  So, some testing is warranted.

ThinkOpenly:
The debuginfo packages are not usable by the `perf` command, since
for a library `/opt/at{version}/lib64/libA.so.0`, perf looks for
`/opt/at{version}/lib64/.debug/libA-{libversion}.so`, resulting in:
```
$ perf probe -v -x /opt/at13.0/lib64/libpthread.so.0 --add 'pthread_mutex_trylock%return $retval'
[...]
Could not open debuginfo. Try to use symbols.
[...]
```

AT currently creates symlinks for
`/opt/at{version}/lib64/.debug/libA-{libversion}.so.debug`
to point to the actual debuginfo file, but not for
`/opt/at{version}/lib64/.debug/libA-{libversion}.so`

Adding the latter, as this patch does, results in:
```
$ perf probe -v -x /opt/at13.0/lib64/libpthread.so.0 --add 'pthread_mutex_trylock%return $retval'
[...]
Open Debuginfo file: /opt/at13.0/lib64/.debug/libpthread-2.30.so
Try to find probe point from debuginfo.
Symbol pthread_mutex_trylock address found : d7a0
Matched function: __pthread_mutex_trylock [66780]
Probe point found: __pthread_mutex_trylock+0
[...]
```

Fixes #965

### DIFF
--- a/configs/14.0/specs/monolithic.spec
+++ b/configs/14.0/specs/monolithic.spec
@@ -318,6 +318,9 @@ for DIR in $(find %{_prefix}/lib/debug/%{_prefix}/lib*/ -type d \
     dest=${DIR#%{_prefix}/lib/debug}
     mkdir -p ${dest}/.debug/
     ln -sf ${DIR}/* ${dest}/.debug/
+    for FILE in $(ls -1 ${DIR}/*.debug 2>/dev/null); do
+        ln -sf ${FILE} ${dest}/.debug/$(basename ${FILE} .debug)
+    done
 done
 
 #---------------------------------------------------
@@ -329,6 +332,9 @@ for DIR in $(find %{_prefix}/lib/debug/%{_prefix}/lib*/ -type d \
     dest=${DIR#%{_prefix}/lib/debug}
     mkdir -p ${dest}/.debug/
     ln -sf ${DIR}/* ${dest}/.debug/
+    for FILE in $(ls -1 ${DIR}/*.debug 2>/dev/null); do
+        ln -sf ${FILE} ${dest}/.debug/$(basename ${FILE} .debug)
+    done
 done
 
 #---------------------------------------------------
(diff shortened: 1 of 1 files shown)


ThinkOpenly:
This leaves the new symlinks around after uninstall. Theh existing code did the same, but let's clean that up...

Hold off reviewing this until I fix that up.

tuliom:
@ThinkOpenly Would it be possible to add a new FVTR testcase to validate this doesn't break in the future?

ThinkOpenly:
@tuliom I don't think so, as all the pieces aren't in place until after the packages are _installed_.

PR #1562, opened 2020-07-22
Disable nonnull check during SPHDE build for AT next

rzinsly:
Until sphde/sphde#53 is fixed we need
to disable "-Werror=nonnull" for SPHDE in order to build it with
GCC 11.


rzinsly:
With #1594 we don't need this anymore.

PR #1570, opened 2020-07-25
Update glibc on AT next

Cherbot:
Bump to revision 7ebd114211dc


er-1:
This fails on Debian while making an AT deb package:
```
perl debian/dh_strip --dest=/home/tcbot/bot-slave/at_opt/at-next-15.0-0-alpha --exclude=getconf --dbg-package=advance-toolchain-at15.0-alpha-devel-dbg -p advance-toolchain-at15.0-alpha-devel
strip: debian/advance-toolchain-at15.0-alpha-devel/home/tcbot/bot-slave/at_opt/at-next-15.0-0-alpha/lib64/libc.a(ppc-mcount.o): unsupported relocation type 0x74
strip: ppc-mcount.o: bad value
dh_strip: strip --strip-debug --remove-section=.comment --remove-section=.note --enable-deterministic-archives debian/advance-toolchain-at15.0-alpha-devel/home/tcbot/bot-slave/at_opt/at-next-15.0-0-alpha/lib64/libc.a returned exit code 1
```

mscastanho:
@tuliom I suspect  by:
```
commit 7c7bcf3634e44cf7e001aaa302138c1ee0e58f8c
Author: Tulio Magno Quites Machado Filho <tuliom@linux.ibm.com>
Date:   Fri Jul 10 19:41:06 2020 -0300

    powerpc64: Fix calls when r2 is not used [BZ #26173]
    
    Teach the linker that __mcount_internal, __sigjmp_save_symbol,
    __syscall_error and __GI_exit do not use r2, so that it does not need to
    recover r2 after the call.
    
    Test at configure time if the assembler supports @notoc and define
    USE_PPC64_NOTOC.
```
According to the commit, `@notoc` is available starting with binutils 2.31. The command that failed used the machine's `strip`, which is of version 2.31.1, so it should have recognized that...

mscastanho:
Looks like the new NOTOC relocation was added to binutils by commit `05d0e962f08`, which `git for-each-ref --contains 05d0e962f08af24` shows first showed up on binutils 2.32.

I noticed that in the packaging process `dh_strip` uses the server's strip, not the one from the binutils we built, so the server's binutils should be >= 2.32 in order for this pass.

ThinkOpenly:
> I noticed that in the packaging process `dh_strip` uses the server's strip, not the one from the binutils we built, so the server's binutils should be >= 2.32 in order for this pass.

Should we be using the AT-built `strip` instead?

mscastanho:
@ThinkOpenly We could for native builds, but that wouldn't work for cross builds.

mscastanho:
After inspecting the packaging process for both DEB and RPM files, looks like DEB files are stripped using `strip` from binutils (see [dh_strip](https://salsa.debian.org/debian/debhelper/-/blob/master/dh_strip) from debhelper), while RPM packages are stripped using `eu-strip` from elfutils (see[ find-debuginfo.sh](https://github.com/rpm-software-management/rpm/blob/master/scripts/find-debuginfo.sh#L257) from rpm-build). This may explain why it fails only on Debian 10 and not on SLES15 and RHEL8, even though all have older versions of binutils.

mscastanho:
Looks like we only strip binaries from native builds, so using the strip we just built might be an option. The following code snippets are only available on the corresponding files for the native builds:

For .deb:
https://github.com/advancetoolchain/advance-toolchain/blob/c9ecea7b65e4ef554a814274e5e68588bd6ac537/configs/14.0/deb/monolithic/rules#L46-L51

For .rpm:
https://github.com/advancetoolchain/advance-toolchain/blob/c9ecea7b65e4ef554a814274e5e68588bd6ac537/configs/14.0/specs/monolithic.spec#L183-L194

PR #1550, opened 2020-07-18
Make debuginfo usable by `perf`

ThinkOpenly:
The debuginfo packages are not usable by the `perf` command, since
for a library `/opt/at{version}/lib64/libA.so.0`, perf looks for
`/opt/at{version}/lib64/.debug/libA-{libversion}.so`, resulting in:
```
$ perf probe -v -x /opt/at13.0/lib64/libpthread.so.0 --add 'pthread_mutex_trylock%return $retval'
[...]
Could not open debuginfo. Try to use symbols.
[...]
```

AT currently creates symlinks for
`/opt/at{version}/lib64/.debug/libA-{libversion}.so.debug`
to point to the actual debuginfo file, but not for
`/opt/at{version}/lib64/.debug/libA-{libversion}.so`

Adding the latter, as this patch does, results in:
```
$ perf probe -v -x /opt/at13.0/lib64/libpthread.so.0 --add 'pthread_mutex_trylock%return $retval'
[...]
Open Debuginfo file: /opt/at13.0/lib64/.debug/libpthread-2.30.so
Try to find probe point from debuginfo.
Symbol pthread_mutex_trylock address found : d7a0
Matched function: __pthread_mutex_trylock [66780]
Probe point found: __pthread_mutex_trylock+0
[...]
```

Also added some `postun` scripts, as currently the uninstall of debuginfo packages would leave behind broken symlinks.

Fixes #965

Signed-off-by: Paul A. Clarke <pc@us.ibm.com>

PR #1590, opened 2020-07-27
Remove last vestiges of OProfile

ThinkOpenly:
OProfile has been removed from AT14, but some mentions remain.
Remove those.
- Package descriptions (required replacement of symlinks with files)
- Release Notes license summary
- Creation of "oprofile" user at install time
- Removing the "binutils_3" target from `binutils.mk`, per the leading comment:
  ```
  # The 3rd binutils build is only necessary when targeting a 64-bit environment
  # and building OProfile for 32-bit.
  ```

### DIFF
--- a/configs/14.0/deb/monolithic/control
+++ b/configs/14.0/deb/monolithic/control
@@ -22,7 +22,7 @@ Architecture: ppc64el
 Pre-Depends: advance-toolchain-__AT_MAJOR_INTERNAL__-runtime (= __AT_FULL_VER__)
 Description: Advance Toolchain
  The advance toolchain is a self contained toolchain which provides preview
- toolchain functionality in GCC, binutils, GLIBC, GDB, Valgrind, and OProfile.
+ toolchain functionality in GCC, binutils, GLIBC, GDB, and Valgrind.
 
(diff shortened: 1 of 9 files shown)


ThinkOpenly:
@mscastanho, thank you for the review!

> I think you edited the wrong file by mistake by following the symlink for configs/14.0/packages/binutils/binutils.mk, which ended up editing the file for AT 6.0. I don't believe we should change the binutils build for older AT versions. You should undo the symlink and edit just the file for AT 14.0 like you did for the others.

That was indeed my mistake.  Fixed and repushed.

> There are still a few comments like the one below referencing Oprofile on configs/14.0/specs/monolithic.spec.

These changes are actually sitting in pull #1550 (for good or bad, but I got there first), so I decided not to include them here.

ThinkOpenly:
> AT 14.0 SLES15 x86_64 20
ck_binaries: x86_64:  Checking if some binaries are available...
Error:  File  /home/tcbot/bot-slave/at_opt/at14.0-0-alpha2/bin/powerpc64le-linux-gnu-gdbserver is missing.

This is already fixed by #1565.

Merging.

PR #1600, opened 2020-07-31
Use AT's strip when generating deb packages

mscastanho:
In the process of generating deb packages for AT, we use dh_strip from
debhelper project to strip our binaries. This script uses the system's
strip, which is almost guaranteed to be an older version than the one we
just built for AT.

Since the AT binaries are generated using a newer binutils, they might
contain symbols the system's strip may not recognize, leading to
problems like:

```
strip: debian/<long-path>/opt/at-next-15.0-0-alpha/lib64/libc.a(ppc-mcount.o):
        unsupported relocation type 0x74
strip: ppc-mcount.o: bad value
```

To avoid this, we can force dh_strip to use the strip binary we just
built, which matches the binutils version used when generating the
binaries. Since we already patch dh_strip before using it, we just have
to update that patch and modify it before applying to use the correct AT
bin dir for the current build.

This commit only applies this change to supported deb-based distros
(Ubuntu and Debian), starting from AT 14.0. The way the patch is
applied should guarantee there are no changes to builds for previous
versions/distros. rpm-based distros are also not modified at this
moment, since they use eu-strip instead, which is not provided by AT.

Lastly, we only strip binaries when doing a native build (see
configs/<config>/deb/{monolithic,monolithic_cross}/rules), so the
changes in this commit should not introduce any changes to cross builds.

---

Closes #1570


mscastanho:
I noted that we also have dh_strip patches for Ubuntu xenial, trusty on `configs/14.0/deb/monolithic`. I didn't update those since this change should only apply to AT >= 14.0 and those are not supported distros for these versions.

configs/14.0/deb/monolithic/bionic_dh_change_dest.patch
```diff
@@ -1 +0,0 @@
-../../../13.0/deb/monolithic/bionic_dh_change_dest.patch
```
ThinkOpenly, line 1: Can you put the link-breaking in a separate commit, so the actual changes are easier to review?
(Exactly what I _didn't_ do in a recent commit of my own. ;-)

configs/next/packages/glibc/sources
```diff
@@ -21,7 +21,7 @@
 
 ATSRC_PACKAGE_NAME="GNU C Library"
 ATSRC_PACKAGE_VER=2.31.9000
-ATSRC_PACKAGE_REV=08538f360f14
+ATSRC_PACKAGE_REV=7ebd114211dc
```
ThinkOpenly, line 24: Did you intend to include this change in this pull request?

scripts/utilities/pkg_build_deb.sh
```diff
@@ -103,7 +103,14 @@ if [[ "${cross_build}" != "yes" ]]; then
...
 		pushd ${deb_d} > /dev/null
 		cp /usr/bin/dh_strip .
-		patch -p1 < ${at_supported_distros}_dh_change_dest.patch
+
+		# For some distros, make sure we also use 'strip' and 'objcopy' from
+		# the binutils we just built instead of the ones from the system. This
+		# specific change will only be applied to distros with a reference to
+		# __AT_BIN__ in their *dh_change_dest.patch file.
+		patch -p1 < <(sed -e "s/__AT_BIN__/${at_dest//\//\\/}\/bin\//g" \
+				${at_supported_distros}_dh_change_dest.patch)
```
ThinkOpenly, line 112: I'll profess my ignorance... :-)  Can you explain what this does:
```
sed -e "s/__AT_BIN__/${at_dest//\//\\/}\/bin\//g"
```
The slash-slash-backslash-slash-slash-backslash-backslash-slash within `${}`, specifically.

configs/14.0/deb/monolithic/bionic_dh_change_dest.patch
mscastanho: @ThinkOpenly that is exactly what 9dff474a91 does. The script for Ubuntu20 (focal) was already a hard file, so I only had to break the links for the other two. This commit breaks the link and just copies the contents without any changes.

configs/next/packages/glibc/sources
mscastanho: @ThinkOpenly Yes. This was the change that actually brought up the whole problem. See #1570. So I added it here.

scripts/utilities/pkg_build_deb.sh
mscastanho: @ThinkOpenly This pattern was used somewhere else in the file, actually. I admit it's hard to understand. `//` replaces any ocurrences of a substring in a bash string. So this replaces any `/` in `$at_dest` with `\/` (back slash + forward slash), so `sed` can correctly print the `/` in the path when replacing the string in the file. Otherwise it would parse `/` as an expression separator. Summary: espace all `/` for `sed`.

So in the end, all that magic is used to replace `__AT_BIN__` in the file with `<path-to-AT-dir>/bin/`

configs/14.0/deb/monolithic/bionic_dh_change_dest.patch
ThinkOpenly: Sorry, you are correct, of course. I was looking at the net changes, not the individual commits. Thanks.

configs/next/packages/glibc/sources
ThinkOpenly: I guess it seems a little out of place in this pull request, but not a big deal.

ThinkOpenly:
Do we need a separate issue for RPM packages?

PR #1613, opened 2020-08-07
Update priority entry of deb packages

### LINKED ISSUE #1564: Use priority optional instead of extra for deb packages
All our deb packages use `extra` as the priority for the AT source files, like in:
https://github.com/advancetoolchain/advance-toolchain/blob/7c94f8029a47af021348ed0a977d5e71f5ced3e1/configs/14.0/deb/monolithic/control#L14-L20

But according to the [Debian Policy Manual](https://www.debian.org/doc/debian-policy/ch-archive.html#s-priorities) this value of priority is deprecated and `optional` should be used instead:

```
extra

    This priority is deprecated. Use the optional priority instead. This
    priority should be treated as equivalent to optional.

    The extra priority was previously used for packages that conflicted with
    other packages and packages that were only likely to be useful to people
    with specialized requirements. However, this distinction was somewhat
    arbitrary, not consistently followed, and not useful enough to warrant the
    maintenance effort.
```

er-1:
Replaced deprecated value, `extra`, of the priority entry in control files.
According to [Debian Policy Manual](https://www.debian.org/doc/debian-policy/ch-archive.html#s-priorities), the `optional` value should be used instead of `extra`.

Fix #1564

### DIFF
--- a/configs/12.0/deb/monolithic/control
+++ b/configs/12.0/deb/monolithic/control
@@ -1,4 +1,4 @@
-# Copyright 2017 IBM Corporation
+# Copyright 2020 IBM Corporation
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


ThinkOpenly:
The changes all look good, but why break the links from `12.0` back to `11.0`?

er-1:
> The changes all look good, but why break the links from `12.0` back to `11.0`?

To apply the change on supported versions only. Not sure it's a good explanation.

PR #1614, opened 2020-08-07
Improve rules files

### LINKED ISSUE #1599: Warnings on pkg_build_deb.sh
While inspecting some logs from an AT next Debian 10 build, I found the following warnings originating from `pkg_build_deb.sh`:

```
$ tail -n+111 _deb-03_pkg_build_deb.log 
dh_installdirs
dh_installdocs
dh_installchangelogs
dh_installchangelogs: warning:     debian/changelog(l2): badly formatted heading line
LINE: #
dh_installchangelogs: warning:     debian/changelog(l6): badly formatted heading line
LINE: #
dh_installchangelogs: warning:     debian/changelog(l8): badly formatted heading line
LINE: #
dh_installchangelogs: warning:     debian/changelog(l14): badly formatted heading line
LINE: #
dh_systemd_enable
dh_install --fail-missing
dh_install: Please use dh_missing --list-missing/--fail-missing instead
dh_install: This feature will be removed in compat 12.
dh_systemd_start
```

The man pages for debhelper states the following about `--fail-missing` change on v12:
```
    Changes from v11 are:
    [...]
    The --list-missing and --fail-missing in dh_install has been removed. Please use dh_missing and its corresponding options, which can also see the files installed by other helpers.
```

The lines above are reached when executing:
https://github.com/advancetoolchain/advance-toolchain/blob/c9ecea7b65e4ef554a814274e5e68588bd6ac537/scripts/utilities/pkg_build_deb.sh#L273

While `debian/rules` will come from the specific config we are building. For example, for AT 14.0 native build:
https://github.com/advancetoolchain/advance-toolchain/blob/c9ecea7b65e4ef554a814274e5e68588bd6ac537/configs/14.0/deb/monolithic/rules#L33-L44

er-1:
Added a call to `dh_missing --fail-missing` after `dh_install`.
This avoids warnings when generation deb packages as `dh_install --fail-missing` has been deprecated.

Fix #1599

### DIFF
--- a/configs/13.0/deb/monolithic/rules
+++ b/configs/13.0/deb/monolithic/rules
@@ -38,7 +38,8 @@ ifeq (__USE_SYSTEMD__, yes)
 endif
 
 # Copy the packages' files.
-	dh_install --fail-missing
+	dh_install
+	dh_missing --fail-missing
 ifeq (__USE_SYSTEMD__, yes)
 	dh_systemd_start
 endif
--- a/configs/14.0/deb/monolithic/rules
+++ b/configs/14.0/deb/monolithic/rules
@@ -38,7 +38,8 @@ ifeq (__USE_SYSTEMD__, yes)
 endif
 
 # Copy the packages' files.
-	dh_install --fail-missing
+	dh_install
+	dh_missing --fail-missing
 ifeq (__USE_SYSTEMD__, yes)
 	dh_systemd_start
 endif
--- a/configs/13.0/deb/monolithic_cross/rules
+++ b/configs/13.0/deb/monolithic_cross/rules
@@ -1,1 +1,66 @@
-../../../12.0/deb/monolithic_cross/rules
+#!/usr/bin/make -f
+#
+# Copyright 2017 IBM Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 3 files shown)


configs/13.0/deb/monolithic/rules
```diff
@@ -38,7 +38,7 @@ ifeq (__USE_SYSTEMD__, yes)
 endif
 
 # Copy the packages' files.
-	dh_install --fail-missing
+	dh_missing --fail-missing
```
ThinkOpenly, line 41: I'm certainly not an expert, but this doesn't seem right.

The comment before the current `dh_install` says "Copy the packages' files.".  My understanding is that is the purpose of `dh_install`.  And `dh_install --fail-missing` will _also_ check for files which were not installed, but presumably should be.  The use of `dh_missing --fail-missing` is just to perform that latter step, correct?  Shouldn't the code be changed to:
```
        dh_install
        dh_missing --fail-missing
```
?

configs/14.0/deb/monolithic/rules
```diff
@@ -38,7 +38,7 @@ ifeq (__USE_SYSTEMD__, yes)
 endif
 
 # Copy the packages' files.
-	dh_install --fail-missing
+	dh_missing --fail-missing
```
ThinkOpenly, line 41: same comment as above, of course.

ThinkOpenly:
Should we also update the copyrights in the newly unlinked files?

configs/13.0/deb/monolithic/rules
er-1: @ThinkOpenly you're right. I updated the patch.

configs/14.0/deb/monolithic/rules
er-1: @ThinkOpenly you're right. I updated the patch.

er-1:
> Should we also update the copyrights in the newly unlinked files?

Done.

PR #1637, opened 2020-08-22
Update binutils on AT next

Cherbot:
Bump to revision 1d4ef7201ffa


mscastanho:
I just glanced over the logs for this error. But the error message was not helpful at all, complaining about invalid chars in temp files with random names.

Closing this to let Cherbot get a new revision so I can take a look again on Monday.

Issue #1670, opened 2020-09-09
Can't install AT 14.0-0 on POWER10


tuliom:
AT 14.0-0 emits the following error when installing on POWER10:

```
Running transaction
  Preparing        :                                                        1/1 
  Running scriptlet: advance-toolchain-at14.0-runtime-14.0-0.ppc64le        1/4 
The system is power10 but must be at least power8 to install this RPM.
error: %prein(advance-toolchain-at14.0-runtime-14.0-0.ppc64le) scriptlet failed, exit status 1
```

er-1:
PR #1671 fixed this. Closing.

PR #1671, opened 2020-09-09
Fix the test for minimum processor during RPM installation

tuliom:
With POWER10, string comparisons do not work as expected between POWER
processors anymore, e.g. "10" < "8".
This patch replaces simple string comparisons with `sort -V`.

Closes issue #1670 .

### DIFF
--- a/configs/11.0/specs/monolithic.spec
+++ b/configs/11.0/specs/monolithic.spec
@@ -214,7 +214,8 @@ echo "enable %{at_ver_alternative}-cachemanager.service" \
 ####################################################
 %pre runtime
 _host_power_arch=$(LD_SHOW_AUXV=1 /bin/true | grep AT_PLATFORM | grep -i power | sed 's/.*power//')
-if [[ "${_host_power_arch}" != "" && "${_host_power_arch}" < "%{_min_power_arch}" ]]; then
+if [[ "${_host_power_arch}" != "" \
+      && "%{_min_power_arch}" != $(echo -e "%{_min_power_arch}\n${_host_power_arch}" | sort -V | head -n 1) ]]; then
     echo "The system is power${_host_power_arch} but must be at least power%{_min_power_arch} to install this RPM."
     exit 1
 fi
--- a/configs/14.0/specs/monolithic.spec
+++ b/configs/14.0/specs/monolithic.spec
@@ -241,7 +241,8 @@ echo "enable %{at_ver_alternative}-cachemanager.service" \
 ####################################################
 %pre runtime
 _host_power_arch=$(LD_SHOW_AUXV=1 /bin/true | grep AT_PLATFORM | grep -i power | sed 's/.*power//')
-if [[ "${_host_power_arch}" != "" && "${_host_power_arch}" < "%{_min_power_arch}" ]]; then
+if [[ "${_host_power_arch}" != "" \
+      && "%{_min_power_arch}" != $(echo -e "%{_min_power_arch}\n${_host_power_arch}" | sort -V | head -n 1) ]]; then
     echo "The system is power${_host_power_arch} but must be at least power%{_min_power_arch} to install this RPM."
     exit 1
 fi
--- a/configs/14.0/specs/monolithic_compat.spec
+++ b/configs/14.0/specs/monolithic_compat.spec
@@ -70,7 +70,8 @@ rm -rf ${RPM_BUILD_ROOT}/%{_prefix}/include/
 ####################################################
 %pre runtime-compat
 _host_power_arch=$(LD_SHOW_AUXV=1 /bin/true | grep AT_PLATFORM | grep -i power | sed 's/.*power//')
-if [[ "${_host_power_arch}" != "" && "${_host_power_arch}" < "%{_min_power_arch}" ]]; then
+if [[ "${_host_power_arch}" != "" \
+	      && "%{_min_power_arch}" != $(echo -e "%{_min_power_arch}\n${_host_power_arch}" | sort -V | head -n 1) ]]; then
     echo "The system is power${_host_power_arch} but must be at least power%{_min_power_arch} to install this RPM."
     exit 1
 fi


er-1 (approved):

configs/11.0/specs/monolithic.spec
```diff
@@ -214,7 +214,8 @@ echo "enable %{at_ver_alternative}-cachemanager.service" \
 ####################################################
 %pre runtime
 _host_power_arch=$(LD_SHOW_AUXV=1 /bin/true | grep AT_PLATFORM | grep -i power | sed 's/.*power//')
-if [[ "${_host_power_arch}" != "" && "${_host_power_arch}" < "%{_min_power_arch}" ]]; then
+if [[ "${_host_power_arch}" != "" \
+      && "%{_min_power_arch}" != $(echo -e "%{_min_power_arch}\n${_host_power_arch}" | sort -V | head -n 1) ]]; then
```
ThinkOpenly, line 218: Could you just change `<` to `-lt` ?

ThinkOpenly (commented):
DEBs don't have this check at all that I can find... is it needed there?  If not, is it needed here?
I propose a simpler change, below.  It does have the disadvantage of failing if the arch is not a simple integer, like "10+" or "10.1".  Though, `sort` may not be reliable in the first case, either.

configs/11.0/specs/monolithic.spec
mscastanho: @ThinkOpenly from my tests, I'm not sure if that would work either
```bash
$ [[ "power8" -lt "power10" ]] && echo true
$ [[ "power10" -lt "power8" ]] && echo true
$ [[ "power10" -eq "power8" ]] && echo true
true
```

mscastanho (approved):

configs/11.0/specs/monolithic.spec
ThinkOpenly: But that's not what's being tested here.  This is a better set of tests:
```
$ [[ "8" -lt "10" ]] && echo true
true
$ [[ "10" -lt "8" ]] && echo true
$
```
(Note that neither variable has "power" as part of its value. The preceding line includes `sed 's/.*power//'`.)

mscastanho: @ThinkOpenly you're right. I was misled by the error message, which has `powerXX` in it. But now I see we are just comparing number strings at this point.

Your solution is nicer to read. I'm fine with either.

tuliom: > DEBs don't have this check at all that I can find... is it needed there? If not, is it needed here?

@ThinkOpenly , this test was added when we dropped support for POWER6, but people continued trying to use AT.
A SIGILL message was difficult to understand, so we had to add this test.
DEBs don't have this test yet because they were created for ppc64le only, which supports POWER >= 8, which we still support.

> I propose a simpler change, below. It does have the disadvantage of failing if the arch is not a simple integer, like "10+" or "10.1".

Historically, we have used, where N is a positive integer:
 - powerN
 - powerN+
 - powerNx

If we endup having a non-decimal suffix, the `-lt` operator returns an error, e.g.:
```
$ [[ "10x" -lt "10" ]] && echo true
bash: [[: 10x: value too great for base (error token is "10x")
```

> Though, sort may not be reliable in the first case, either.

Why isn't it reliable?

```
$ echo -e "10x\n10+\n10\n8" | sort -V
8
10
10x
10+
```

ThinkOpenly: > > DEBs don't have this check at all that I can find... is it needed there? If not, is it needed here?
> 
> @ThinkOpenly , this test was added when we dropped support for POWER6, but people continued trying to use AT.
> A SIGILL message was difficult to understand, so we had to add this test.
> DEBs don't have this test yet because they were created for ppc64le only, which supports POWER >= 8, which we still support.

So, we can remove this test for AT >= 12?  One of the changes is in a config file for AT14.

> > I propose a simpler change, below. It does have the disadvantage of failing if the arch is not a simple integer, like "10+" or "10.1".
> 
> Historically, we have used, where N is a positive integer:
> 
> * powerN
> * powerN+
> * powerNx
> 
> If we endup having a non-decimal suffix, the `-lt` operator returns an error, e.g.:
> 
> ```
> $ [[ "10x" -lt "10" ]] && echo true
> bash: [[: 10x: value too great for base (error token is "10x")
> ```

Right, that's what I was worried about when arch is not a simple integer.

> > Though, sort may not be reliable in the first case, either.
> 
> Why isn't it reliable?
> 
> ```
> $ echo -e "10x\n10+\n10\n8" | sort -V
> 8
> 10
> 10x
> 10+
> ```

It won't be reliable if we ever set the minimum to "10x" _and_ there's a "10+" (unlikely), or if we ever do some other suffix:
```
$ echo -e "10+\n10x\n10\n10.1" | sort -V
10
10x
10+
10.1
```
...which might stray into bikeshedding.

`sort|head` works, and is a bit more robust.

tuliom: > So, we can remove this test for AT >= 12? One of the changes is in a config file for AT14.

@ThinkOpenly We could remove it for AT >= 13.0.  Notice that AT 12.0 still supports RHEL 7 and ppc64.
If we remove the test in those versions, we may need to re-add it if/when we drop support for POWER8.

ThinkOpenly: > > So, we can remove this test for AT >= 12? One of the changes is in a config file for AT14.
> 
> @ThinkOpenly We could remove it for AT >= 13.0. Notice that AT 12.0 still supports RHEL 7 and ppc64.

Yes, that's what I meant, sorry.  AT >= 13.

> If we remove the test in those versions, we may need to re-add it if/when we drop support for POWER8.

Ah, if we drop support of a processor version before a supported distro does... good point.
We'd need to add tests in the DEB scripts, then, too.

ThinkOpenly (approved):

PR #1672, opened 2020-09-11
Update valgrind on AT 14.0

er-1:
Bump to version 3.16.1
Bump to revision 24f63fd435e7

Also, on AT next, replace version with 'master' keyword as AT next tracks the master branch of valgrind.


er-1:
The 2 failures on Debian 10 x86_64 are due to lack of disk space on the tcbot worker.

PR #1675, opened 2020-09-12
Update gcc on AT next

Cherbot:
Bump to revision ac35c090821f

Issue #1667, opened 2020-09-08
GCC on AT next failing with internal compiler error (ICE)


mscastanho:
I'm creating this issue so we can track this bug while it has not been fixed upstream, to avoid having info about it scattered between many failed GCC bump PRs.

Starting with #1639, AT next build is failing with the following error:

Failed when building power10-tuned GCC and OpenSSL. In both cases I see similar errors:

- GCC
```
during RTL pass: dse1
In file included from /home/mscastanho/~/AT/next/tmp/build/at15.0-0-alpha.redhat-8_ppc64le_ppc64le/sources/gcc/libsanitizer/asan/asan_interceptors.cpp:172:
/home/mscastanho/~/AT/next/tmp/build/at15.0-0-alpha.redhat-8_ppc64le_ppc64le/sources/gcc/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc: In function ‘__sanitizer::__sanitizer_FILE* __interceptor_fopencookie(void*, const char*, __sanitizer::__sanitizer_cookie_io_functions_t)’:
/home/mscastanho/~/AT/next/tmp/build/at15.0-0-alpha.redhat-8_ppc64le_ppc64le/sources/gcc/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:6216:1: internal compiler error: in convert_mode_scalar, at expr.c:412
 6216 | }
      | ^
0x7fff8f729edf generic_start_main
        ../csu/libc-start.c:314
0x7fff8f72a0bf __libc_start_main
        ../sysdeps/unix/sysv/linux/powerpc/libc-start.c:98
```

- OpenSSL
```
during RTL pass: dse1
crypto/blake2/blake2b.c: In function 'blake2b_compress':
crypto/blake2/blake2b.c:202:1: internal compiler error: in convert_mode_scalar, at expr.c:412
  202 | }
      | ^
0x1094808b convert_mode_scalar
        /home/mscastanho/~/AT/next/tmp/build/at15.0-0-alpha.redhat-8_ppc64le_ppc64le/sources/gcc/gcc/expr.c:412
0x1094150b convert_modes(machine_mode, machine_mode, rtx_def*, int)
        /home/mscastanho/~/AT/next/tmp/build/at15.0-0-alpha.redhat-8_ppc64le_ppc64le/sources/gcc/gcc/expr.c:737
0x11088d67 extract_low_bits(machine_mode, machine_mode, rtx_def*)
[...]

PR #1659, opened 2020-09-05
Update glibc on AT next

Cherbot:
Bump to revision 04bba1e5d84b

PR #1657, opened 2020-09-05
Update gcc on AT next

Cherbot:
Bump to revision bec05c98b94f

PR #1690, opened 2020-09-16
Bump glibc + gcc on AT next

mscastanho:
We have a chain of failures that requires updating a few packages in an specific order:

1) glibc update is causing binutils to fail because of the deprecation of `mallinfo` (see comments in #1659)
2) gcc update is causing glibc to fail because of new warning (see #1657 and #1675)

The fix for (2) is already upstream on glibc, but we didn't get it because of (1).
Also, we have not updated gcc in a while even before that because of #1667, which already has a temp fix upstream.

So this PR does the following:
1) Suppresses `-Werror=deprecated-declarations` on binutils stage 2 _temporarily_ (see #1691)
     This way we can avoid the error and bump glibc
2) Updates glibc sources
     To get the fix for the `-Werror=stringop-overread` error
3) Updates gcc sources


mscastanho:
gold is written in C++, so the flag to avoid the warning had to go to `CXXFLAGS`, instead of `CFLAGS`.

Still, all the cross targets failed because of a timeout when building OpenSSL. Exactly on the same file we saw the POImode error in #1667.

mscastanho:
Another failure (besides the one on OpenSSL): 
- GCC power10 tuned, while linking libgo
```
/home/tcbot/bot-slave/at_opt/at-next-15.0-0-alpha/powerpc64le-linux-gnu/bin/ld.gold: error: runtime/.libs/go-cdiv.o: failed to match split-stack sequence at section 1 offset 0
/home/tcbot/bot-slave/at_opt/at-next-15.0-0-alpha/powerpc64le-linux-gnu/bin/ld.gold: error: runtime/.libs/go-cdiv.o: failed to match split-stack sequence at section 1 offset 70
/home/tcbot/bot-slave/at_opt/at-next-15.0-0-alpha/powerpc64le-linux-gnu/bin/ld.gold: error: runtime/.libs/runtime_c.o: failed to match split-stack sequence at section 1 offset 310
/home/tcbot/bot-slave/at_opt/at-next-15.0-0-alpha/powerpc64le-linux-gnu/bin/ld.gold: error: reflect/.libs/makefunc_ffi_c.o: failed to match split-stack sequence at section 1 offset 1a0
```

mscastanho:
There is a GCC bug report related to that issue: [BZ97107](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=97107).

mscastanho:
The timeout issue when building OpenSSL should be fixed now since [this commit](https://gcc.gnu.org/git/gitweb.cgi?p=gcc.git;h=080ddd8a81ea39e92542fa7d7e2d7d27ee741710).

And the fix for the gold issue has a patch pending on gcc-patches: [link](https://gcc.gnu.org/pipermail/gcc-patches/2020-September/554453.html).

I'm just waiting for the last fix to be merged and I'll update this PR to get all working.

PR #1695, opened 2020-09-19
Update glibc on AT next

Cherbot:
Bump to revision cdf645427d17


mscastanho:
All needed fixes to update glibc are being included in #1690.

PR #1698, opened 2020-09-23
Remove most pre-dependencies on deb packages

tuliom:
5 of our deb packages have a Pre-Depends relationship because their
postinst(configure) execution depend on the availability of ldconfig.
However, many other packages had a Pre-Depends relationship without
having a pre-configure dependency.

### DIFF
--- a/configs/next/deb/monolithic/control
+++ b/configs/next/deb/monolithic/control
@@ -1,1 +1,180 @@
-../../../14.0/deb/monolithic/control
+# Copyright 2017 IBM Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+Source: advance-toolchain-__AT_MAJOR_INTERNAL__
+Section: devel
+Priority: optional
+Maintainer: Advance Toolchain
+
+Package: advance-toolchain-runtime
+Architecture: ppc64el
+Pre-Depends: advance-toolchain-__AT_MAJOR_INTERNAL__-runtime (= __AT_FULL_VER__)
+Description: Advance Toolchain
+ The advance toolchain is a self contained toolchain which provides preview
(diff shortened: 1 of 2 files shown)


ThinkOpenly (approved):

tuliom:
```
dh_testdir: error: depends-field appears twice in the same stanza of debian/control. First time on line 43, second time: 44
```
That needs to be fixed.

tuliom:
I updated this PR with 2 changes:
1. Fixed (hopefully!) the previous build failures.
2. Added a new patch that removes the rest of the `Pre-Depends` entries.

I still believe this is still draft material because:
1. We need to decide if we're going to change previous versions.
2. Test. I didn't test my changes and these patches do require pedantic tests.

ThinkOpenly (approved):

ThinkOpenly:
> 1. Fixed (hopefully!) the previous build failures.

:+1: 

> 2. Added a new patch that removes the rest of the `Pre-Depends` entries.

:+1:

> I believe this is still draft material because:
> 
> 1. We need to decide if we're going to change previous versions.

I would be OK _not_ changing previous versions, because these types of changes always seem to be more complicated than they appear.

> 2. Test. I didn't test my changes and these patches do require pedantic tests.

I suspect we can be no worse off than without the changes, because it is currently a failure when installing multiple packages simultaneously using `dpkg`.  I further suspect these changes will only impact that scenario, and for the better.  These changes only obviously impact _installation_, so latent bugs at package removal seem unlikely.  (But, my suspicions can't replace testing!  :-)

rzinsly:
> > 1. We need to decide if we're going to change previous versions.
> 
> I would be OK _not_ changing previous versions, because these types of changes always seem to be more complicated than they appear.
> 

I agree.

Seems we need changes in ck_provides as well.

tuliom:
@ThinkOpenly , @rzinsly , I modified the patches to change only AT next and fixed the errors on ck_provides.
I'm also removing the draft status of this PR. I think we can merge this after it gets enough approvals.

tuliom:
So far, the failures on AT <= 12.0 are related to Python.
Waiting on PR #1749 in order to progress with this.

ThinkOpenly:
> @ThinkOpenly , @rzinsly , I modified the patches to change only AT next and fixed the errors on ck_provides.
> I'm also removing the draft status of this PR. I think we can merge this after it gets enough approvals.

Still LGTM, and my approval is still in place.

rzinsly (approved):
LGTM

tuliom:
I rebased the PR on top of the master branch hoping that all the tests pass this time.

tuliom:
My last rebase did not have all the fixes on AT 10.0.  Trying again...

tuliom:
33 successful checks and 2 approvals.
I think this is ready to be merged.
Thanks!

PR #1703, opened 2020-09-26
Update gcc on AT next

Cherbot:
Bump to revision d4a906e7b51f


mscastanho:
Failed to apply patches to libffi. I think we don't need that anymore.
I'm preparing a PR to fix this.

PR #1713, opened 2020-09-28
Remove libffi patch to GCC

mscastanho:
We were applying an off-tree patch to fix an issue on libffi on Power10.
However, this issue is now fixed on GCC upstream by:

    master:
    fff56af6421 Re: [RS6000] Power10 libffi fixes

    ibm/gcc-10-branch:
    71c83e108de [RS6000] Power10 libffi fixes

Our patch is not needed anymore, so remove it.

Closes #1703
Closes #1710


mscastanho:
Also bumping AT 14.0 and AT next

mscastanho:
The pending builds are bogus. They will never end. They got replaced by others (which have already passed).

lucmaga (approved):

tuliom (approved):

PR #1717, opened 2020-10-03
Update gcc on AT next

Cherbot:
Bump to revision b0b9b8f02a1c


mscastanho:
Failed on GCC stage 2. Looks like some failure while building documentation:
```
  47 genhooks: No place specified to document hook TARGET_ASM_OPEN_PAREN                                                                                                                                                                                     
  48                                                                                                                                                                                                                                                         
  49 2863 translated messages, 7383 fuzzy translations, 3754 untranslated messages.                                                                                                                                                                          
  50 echo timestamp > s-target-hooks-def-h                                                                                                                                                                                                                   
  51 echo timestamp > s-common-target-hooks-def-h                                                                                                                                                                                                            
  52 echo timestamp > s-c-target-hooks-def-h                                                                                                                                                                                                                 
  53 make[4]: *** [Makefile:2560: s-tm-texi] Error 1
```
Trying a newer revision...

mscastanho:
Latest revision has the same issue. Trying to reproduce it on upstream GCC...

mscastanho:
I was not able to reproduce it with GCC upstream, but I have more info.

I found out that the first while loop in genhooks was not finding any `@hook` strings, that's why we are seeing the error. The hash table is not being filled in the first place, so when genhooks tries to process `TARGET_ASM_OPEN_PAREN`, it fails. This would happen with any hook, it just happens that this is the first one to be processed and thus the first one to trigger the issue.

I was able to create a reproducer. The issue happens when we use AT's dynamic linker. Still investigating...

```
$ cat ./mimic-genhooks.c
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

char buf[1000];

int main(int argc, char** argv){
  FILE *f;

  if(argc != 2)
    return -1;

  /* Enter all the start hooks in start_hooks.  */
  f = fopen (argv[1], "r");
  if (!f)
      exit(1);
  while (fscanf (f, "%*[^@]"), buf[0] = '\0',
         fscanf (f, "@%5[^ \n]", buf) != EOF)
    {
      if (strcmp (buf, "hook") != 0)
        continue;
      buf[0] = '\0';
      fscanf (f, "%999s", buf);
      printf("hook name: %s\n", buf);
    }
  fclose (f);

  return 0;
}
$ /home/mscastanho/AT/broken-gcc-genhooks/opt/at-next-15.0-0-alpha/bin/g++ -Wl,--dynamic-linker=/home/mscastanho/AT/broken-gcc-genhooks/o
pt/at-next-15.0-0-alpha/lib64/ld64.so.2  -o mimic-genhooks /home/mscastanho/mimic-genhooks.c
$ ./mimic-genhooks /home/mscastanho/AT/broken-gcc-genhooks/build/at15.0-0-alpha.redhat-8_ppc64le_ppc64le/sources/gcc/gcc/doc/tm.texi.in
$ /home/mscastanho/AT/broken-gcc-genhooks/opt/at-next-15.0-0-alpha/bin/g++ -o mimic-genhooks /home/mscastanho/mimic-genhooks.c
$ ./mimic-genhooks /home/mscastanho/AT/broken-gcc-genhooks/build/at15.0-0-alpha.redhat-8_ppc64le_ppc64le/sources/gcc/gcc/doc/tm.texi.in
hook name: TARGET_ALWAYS_STRIP_DOTDOT
hook name: TARGET_DEFAULT_TARGET_FLAGS
hook name: TARGET_HANDLE_OPTION
hook name: TARGET_HANDLE_C_OPTION
<snip>
```

PR #1535, opened 2020-07-04
Update libpfm on AT 12.0

Cherbot:
Bump to revision 5a623727cf71

PR #1724, opened 2020-10-06
Update libpfm on AT 12.0 and fix build issue

mscastanho:
Bump to revision fa84c27b6057

This bump includes some changes to the upstream project's Makefile (libpfm
commit 6a76d09db) that require us to change our build steps. The hack to replace
the PREFIX value is not necessary anymore, as we can now set that value when
calling make.

Also, BITMODE var was removed upstream(libpfm commit d47158859), so no need to
set it anymore.

Also update the package version.


mscastanho:
This fixes the problems caused by #1535

mscastanho:
Rebased against master to get the fixes for Python builds.

er-1:
Failing while running python test `test_distutils`:
```
test test_distutils failed -- Traceback (most recent call last):
  File "/home/tcbot/bot-slave/at_opt/at12.0/lib64/python3.6/distutils/tests/test_bdist_wininst.py", line 26, in test_get_exe_bytes
    exe_file = cmd.get_exe_bytes()
  File "/home/tcbot/bot-slave/at_opt/at12.0/lib64/python3.6/distutils/command/bdist_wininst.py", line 360, in get_exe_bytes^M
    f = open(filename, "rb")
FileNotFoundError: [Errno 2] No such file or directory: '/home/tcbot/bot-slave/at_opt/at12.0/lib64/python3.6/distutils/command/wininst-14.0.exe'
```

mscastanho:
Rebased against current master to get the fixes from #1749. Hopefully now all tests will pass.

tuliom (changes requested):
The update looks good to me for a "new" AT release.
I was not expecting this kind of update in a stable AT release such as AT 12.0 unless with have a good justification.

In this scenario, we would still use libpfm 4.11.1, but apply the fix on top of that.
Would it be possible?

mscastanho:
@tuliom Updated the PR to use libpfm 4.11.1 instead of a newer revision.

configs/12.0/packages/libpfm/sources
```diff
@@ -20,8 +20,8 @@
 #
 
 ATSRC_PACKAGE_NAME="Libpfm"
-ATSRC_PACKAGE_VER=4.10.1
-ATSRC_PACKAGE_REV=37d4628e37ba
+ATSRC_PACKAGE_VER=4.11.1
```
tuliom, line 23: @mscastanho is this version really correct?

tuliom (changes requested):

configs/12.0/packages/libpfm/sources
mscastanho: @tuliom I though so. I'm a bit confused. Should we keep using 4.10.1 then instead of 4.11.1? From your previous comment I though we should be using 4.11.1, but not a more recent revision.

tuliom: @mscastanho Oh! That was a typo... I meant to say 4.10.1.
The whole point is: why are we modifying this stable code?

The general rule is:
Going from 4.10.0 to 4.10.1 usually means that you're getting only bugfixes.
Going from 4.10.1 to 4.11.1 may also mean you're getting changes that may be incompatible with software built for 4.10. Sometimes we do have good reasons to have this kind of update, but I can't see it now.

mscastanho: @tuliom This PR was meant to fix the issues seem by a Cherbot bump (#1535) which was updating libpfm to revision `5a623727cf71` (well past v4.10.1). Later we even got another automated bump (#1742) which tried to update libpfm to revision `fa84c27b6057` (current head of master).

So was Cherbot wrongly bumping this package in the first place? If that is the case, then this PR might not be needed at all.

er-1: > So was Cherbot wrongly bumping this package in the first place? If that is the case, then this PR might not be needed at all.

If libpfm is going too far for AT 12.0, may be it's time for Cherbot to stop watching libpfm.
As @mscastanho said, sadly this PR won't be needed if we take nothing more than libpfm 4.10.1 on AT 12.0.

mscastanho:
We realized that the libpfm updates done by Cherbot (#1535 and #1742) were mistaken, as we don't have any special reason to update it to a newer version than 4.10.1 (what we already have).  So this PR are not needed.

PR #1727, opened 2020-10-09
Remove GCC stage 2 from AT next

tuliom:
A while ago, both GCC and glibc received changes that allowed to build
glibc early, immediately after the first GCC build.  See the difference
between AT 7.0 and 7.1.
    
This modification allowed AT to remove either GCC's stage 2 or stage 3
and provide the same feature-complete GCC that stage 3 provides.

### DIFF
--- a/configs/next/packages/gcc/gcc.mk
+++ b/configs/next/packages/gcc/gcc.mk
@@ -1,1 +1,77 @@
-../../../14.0/packages/gcc/gcc.mk
+# Copyright 2017 IBM Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+#
+#
+$(eval $(call set_provides,gcc_1,single,cross_yes))
+$(eval $(call set_provides,gcc_2,single,cross_yes))
+$(eval $(call set_provides,gcc_3,single,cross_yes))
+$(eval $(call set_provides,gcc_4,single,cross_yes))
+
+# List of dependencies in order to build the tuned libraries.
+gcc_tuned-deps := $(RCPTS)/gcc_4.rcpt
+# Enable tuned targets
+$(eval $(call provide_tuneds,gcc))
+
+# gcc_4 doesn't require mpc_2 on cross.
+ifeq ($(CROSS_BUILD),no)
+    gcc_4-reqs += $(RCPTS)/mpc_2.rcpt
+endif
+
+
+gcc_1: $(RCPTS)/gcc_1.rcpt
+
+gcc_2: $(RCPTS)/gcc_2.rcpt
+
+gcc_3: $(RCPTS)/gcc_3.rcpt
+
+gcc_4: $(RCPTS)/gcc_4.rcpt
+
+gcc_tuned: $(RCPTS)/gcc_tuned.rcpt
+
+$(RCPTS)/gcc_1.rcpt: $(gcc_1-archdeps)
+	@touch $@
+
+$(RCPTS)/gcc_2.rcpt: $(gcc_2-archdeps)
+	@touch $@
+
+$(RCPTS)/gcc_3.rcpt: $(gcc_3-archdeps)
+	@touch $@
+
+$(RCPTS)/gcc_4.rcpt: $(gcc_4-archdeps)
+	@touch $@
+
+
+$(RCPTS)/gcc_1.a.rcpt: $(RCPTS)/binutils_1.rcpt \
+                       $(RCPTS)/mpc_1.rcpt \
+                       $(RCPTS)/kernel_h.rcpt \
+                       $(RCPTS)/rsync_gcc.rcpt
+	@touch $@
+
+$(RCPTS)/gcc_2.a.rcpt: $(RCPTS)/gcc_1.rcpt \
+                       $(RCPTS)/glibc_1.rcpt \
+                       $(RCPTS)/ldconfig_1.rcpt
+	@touch $@
+
+$(RCPTS)/gcc_3.a.rcpt: $(RCPTS)/ldconfig_1.rcpt \
+                       $(RCPTS)/gcc_2.rcpt
+	@touch $@
+
+$(RCPTS)/gcc_4.a.rcpt: $(RCPTS)/binutils_2.rcpt \
+                       $(RCPTS)/gcc_3.rcpt \
+                       $(gcc_4-reqs)
+	@touch $@
+
+$(RCPTS)/gcc_tuned.rcpt: $(gcc_tuned-archdeps)
+	@touch $@
--- a/configs/next/packages/gcc/gcc.mk
+++ b/configs/next/packages/gcc/gcc.mk
@@ -15,7 +15,6 @@
 #
 #
 $(eval $(call set_provides,gcc_1,single,cross_yes))
-$(eval $(call set_provides,gcc_2,single,cross_yes))
 $(eval $(call set_provides,gcc_3,single,cross_yes))
 $(eval $(call set_provides,gcc_4,single,cross_yes))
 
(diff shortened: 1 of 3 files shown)


configs/next/packages/gcc/stage_3
```diff
@@ -35,6 +35,23 @@ atcfg_pre_hacks() {
 		[[ -h ${Z%\.a}_s.a ]] && rm -f ${Z%\.a}_s.a
 		[[ -h ${Z%\.a}_eh.a ]] && rm -f ${Z%\.a}_eh.a
 	done
+
```
ThinkOpenly: Is the code above still needed?  "Clean static lib hack done on gcc stage 2", now that stage 2 is defunct?

configs/next/packages/gcc/stage_3
```diff
@@ -120,36 +147,86 @@ atcfg_configure() {
...
 # Make build command
 atcfg_make() {
-	if [[ "${cross_build}" == "no" ]]; then
-		PATH=${at_dest}/bin:${PATH} \
-			${SUB_MAKE} all
-	else
-		echo "No make all on gcc_3 cross stage"
-	fi
+	PATH=${at_dest}/bin:${PATH} \
+		${SUB_MAKE} all
```
ThinkOpenly: Is this an unrelated change?

configs/next/packages/gcc/stage_3
```diff
@@ -120,36 +147,86 @@ atcfg_configure() {
...
 }
 
 # Conditional install build command
 atcfg_install() {
-	if [[ "${cross_build}" == "no" ]]; then
-		make install -j1 DESTDIR=${install_place}
-	else
-		echo "No make install on gcc_3 cross stage"
-	fi
+	make install -j1 DESTDIR=${install_place}
```
ThinkOpenly: Is this an unrelated change?

configs/next/packages/gcc/stage_3
```diff
@@ -120,36 +147,86 @@ atcfg_configure() {
...
 }
 
 # Conditional post install settings or commands to run
 atcfg_post_install() {
-	if [[ "${cross_build}" == "no" ]]; then
-		# Remove unnecessary python scripts
-		find "${install_place}/${at_dest}/lib" "${install_place}/${at_dest}/lib64" -name "libstdc++.so.*-gdb.py" -print -delete
```
ThinkOpenly: unrelated change? (also tied to the `find` command "added" well below)

ThinkOpenly:
It's great to see this change :1st_place_medal: !

Would it be an imposition to ask for more explanation about why this is now possible?  The commit message is a bit vague:
> See the difference between AT 7.0 and 7.1.
Since that was, what, 7 years ago, why did GCC stage 2 hang around for so long?  :-)

Do we need to do any performance validation?  Or, can we just compare the resulting binaries?  Should they be identical?

It's a little difficult to compare the differences, as `configs/next/packages/gcc/stage_2` was left as a symlink and removed, yet some of its content was moved up to `stage_3`.  But, I think I got the gist ... reviewed separately.

ThinkOpenly:
Any "build time" comparisons? (Can you tell I find this a very welcome change?  :-)

configs/next/packages/gcc/stage_3
tuliom: No.  We didn't use to build gcc_3 on cross compiler. As we remove gcc_2, this is now mandatory.

configs/next/packages/gcc/stage_3
tuliom: Likewise, regarding the cross compiler.

configs/next/packages/gcc/stage_3
tuliom: Likewise for the cross compiler.
We have to run this command both on native and cross builds.

configs/next/packages/gcc/stage_3
tuliom: No it isn't needed anymore.  I missed this.

tuliom:
> Would it be an imposition to ask for more explanation about why this is now possible?

@ThinkOpenly Until AT 7.0, we used to have the following build order:
```
...
gcc_1
glibc_h (just for headers)
gcc_2 (configured with the headers from glibc_h, did not have all features)
glibc_1 (first glibc build, still incomplete)
gcc_3 (feature complete)
glibc_2 (feature complete)
...
```
After changes to gcc and [glibc](https://sourceware.org/git/?p=glibc.git;a=blob;f=NEWS;h=bdbf52bd7c05dd9bb321caa67f69c27173eb7a52;hb=refs/heads/release/2.19/master#l364) (I don't remember the details) we were able to build glibc_1 immediately after gcc_1.
We agreed that was a first step and we had to review GCC builds later, with further changes.

> Since that was, what, 7 years ago, why did GCC stage 2 hang around for so long? :-)

I think we were still too happy with a 4h build time (down from 16h on AT 6.0). :-D

> Do we need to do any performance validation? Or, can we just compare the resulting binaries? Should they be identical?

We could compare the binaries. But I admit I'm confident this is doing the right thing because I think that if it was wrong, gcc_3 would fail.  And because I still remember the discussion on libc-alpha.

> Any "build time" comparisons? (Can you tell I find this a very welcome change? :-)

In my tests, gcc_2 was taking ~40 minutes to complete. That's what I expect to save with this patch.
I was not able to compare total build time on my runs because of the download of the source code.  I'm waiting for the results from the CI in order to compare.

configs/next/packages/gcc/gcc.mk
```diff
@@ -59,13 +53,8 @@ $(RCPTS)/gcc_1.a.rcpt: $(RCPTS)/binutils_1.rcpt \
...
 	@touch $@
 
-$(RCPTS)/gcc_2.a.rcpt: $(RCPTS)/gcc_1.rcpt \
-                       $(RCPTS)/glibc_1.rcpt \
-                       $(RCPTS)/ldconfig_1.rcpt
-	@touch $@
-
 $(RCPTS)/gcc_3.a.rcpt: $(RCPTS)/ldconfig_1.rcpt \
-                       $(RCPTS)/gcc_2.rcpt
+                       $(RCPTS)/glibc_1.rcpt
```
ThinkOpenly: This removes some dependencies (`gcc_1`, `ldconfig_1`). Is that intentional?

ThinkOpenly:
> > Would it be an imposition to ask for more explanation about why this is now possible?
> 
> @ThinkOpenly Until AT 7.0, we used to have the following build order:
> 
> ```
> ...
> gcc_1
> glibc_h (just for headers)
> gcc_2 (configured with the headers from glibc_h, did not have all features)
> glibc_1 (first glibc build, still incomplete)
> gcc_3 (feature complete)
> glibc_2 (feature complete)
> ...
> ```
> 
> After changes to gcc and [glibc](https://sourceware.org/git/?p=glibc.git;a=blob;f=NEWS;h=bdbf52bd7c05dd9bb321caa67f69c27173eb7a52;hb=refs/heads/release/2.19/master#l364) (I don't remember the details) we were able to build glibc_1 immediately after gcc_1.
> We agreed that was a first step and we had to review GCC builds later, with further changes.

Thanks for the information.  Should we add any of the above information into the commit message?

tuliom:
@ThinkOpenly I improved the commit message and updated the PR.

ThinkOpenly:
Just one "unresolved conversation" left, regarding the change in dependencies. (gentle ping)

configs/next/packages/gcc/gcc.mk
mscastanho: @ThinkOpenly note the `ldconfig_1` dependency was already part of gcc_3 (the white line not in the diff). But I have the same question regarding the `gcc_1` dep.

mscastanho (approved):
Taking a look at the changes, looks like we are effectively using the config for stage 2 on stage 3, so we are actually removing stage_3. But in the end it should yield the same result.

In the future, it would be nice to rename our stages to have 1-2-3-opt, instead of 1-3-4-opt. It's less confusing. But no need to include it in this PR.

More importantly, I built AT next on Debian 10 with these changes and compared against a build based on 436b19da32fa416eb4 (right before this change) using
```
diffoscope ~/AT/before-remove-gcc2/opt/at-next-15.0-0-alpha/ ~/AT/remove-gcc2/opt/at-next-15.0-0-alpha/ --exclude-directory-metadata=recursive
```
The flag instructs diffoscope to ignore file timestamps and other metadata. The `recursive` mode is more meaningful when comparing `.deb` or `.tar` files.

I also ran a similar command to compare every `.deb` file generated by the builds. The output was clobbered by the changes in install path, since I configured the builds to not use `/opt` and also used different dirs. But from what I could see, the only differences were the file path prefixes, so all is good.

Any change that reduces build time is a very welcome change IMO. :tada: 
LGTM

configs/next/packages/gcc/gcc.mk
mscastanho: @tuliom Makes sense. The `gcc_1` dep is satisfied implicitly through `glibc_1`, so removing that  really doesn't change anything in the end. Thanks for the explanation!

ThinkOpenly: OK

ThinkOpenly (approved):

PR #1732, opened 2020-10-10
Update glibc on AT next

Cherbot:
Bump to revision 81b83ff61f95


mscastanho:
GCC stage 3 failed with:
```
/home/tcbot/bot-slave/at_opt/at-next-15.0-0-alpha/bin/ld: /home/tcbot/bot-slave/at/src/at15.0-0-alpha.redhat-8_ppc64le_ppc64le/builds/gcc_3/./gcc/liblto_plugin.so: error loading plugin: /lib64/libc.so.6: version `GLIBC_2.33' not found (required by /home/tcbot/bot-slave/at/src/at15.0-0-alpha.redhat-8_ppc64le_ppc64le/builds/gcc_3/./gcc/liblto_plugin.so)
```

mscastanho:
Ok, a little bit more info about what is going on:
During libgcc linking step the linker tries to load `liblto_plugin.so` to apply link time optimization (LTO), which has references to new glibc symbols introduced with the latest bump:
```
$ readelf -s ~/AT/broken-glibc-233-symbol/build/at15.0-0-alpha.redhat-8_ppc64le_ppc64le/builds/gcc_3/gcc/liblto_plugin.so | grep GLIBC_2.33
    34: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND lstat@GLIBC_2.33 (3)         [<localentry>: 8] 
    41: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND stat@GLIBC_2.33 (3)  [<localentry>: 8] 
   450: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND lstat@GLIBC_2.33     [<localentry>: 8] 
   464: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND stat@GLIBC_2.33      [<localentry>: 8]
```
The linker is failing to find those symbols, because it is using the system's dynamic linker instead of AT's:
```
$ file /home/mscastanho/AT/broken-glibc-233-symbol/opt/at-next-15.0-0-alpha/bin/ld 
/home/mscastanho/AT/broken-glibc-233-symbol/opt/at-next-15.0-0-alpha/bin/ld: ELF 64-bit LSB executable, 64-bit PowerPC or cisco 7500, version 1 (SYSV), dynamically linked, interpreter /lib64/ld64.so.2, for GNU/Linux 3.10.0, BuildID[sha1]=fc441382d6972681500930eff14fd003feed570e, with debug_info, not stripped
```

But AFAIU, there is no way the linker could use AT's dynamic linker at this point, since it is still the linker built on binutils stage 1, which runs before glibc_1 and even gcc_1.

Am I missing something?

tuliom:
I think you're correct.
We can only enable enable LTO optimization after binutils_2.  In other words: only in gcc_4.

rff:
This is very similar to a problem we had with gcc_2, @mscastanho take a look at this PRs (https://github.com/advancetoolchain/advance-toolchain/pull/1149, https://github.com/advancetoolchain/advance-toolchain/pull/1160) and see if it is of any help.

PR #1749, opened 2020-10-14
Disable more Python tests on AT <= 12.0

tuliom:
Some Python tests have been unreliable on AT <= 12.0.

test_distutils is known to fail when run from the install directory.

### DIFF
--- a/fvtr/python/python.exp
+++ b/fvtr/python/python.exp
@@ -100,15 +100,21 @@ if { [lindex $pyverl 0] > 3
 	append common_exclude_tests " test_dtrace"
 }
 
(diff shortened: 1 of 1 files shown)


tuliom:
I expect python tests to fail on AT <= 11.0 at first.
However, this version of this PR should fix the failures on AT 12.0.
I plan to work slowly in order to get these tests passing.  There are too many test failures right now.

tuliom:
I merged this PR because it's improving the results on AT 11.0 and 12.0.
I hope this will unblock patches affecting only that version.
I plan to continue to work on AT 10.0.

PR #1751, opened 2020-10-15
Disable LTO on GCC stage 3

mscastanho:
This is a fix for the issue seen in #1732. For more info about it, please see the comments on that PR and the commit messages for this PR. I'm also adding the glibc bump from #1732 to make sure the issue is tested and fixed.

Disabling LTO on GCC stage 3 avoids the problem. This is what this PR is doing. But something is still not clear to me.

Despite us [explicitly enabling LTO on GCC stage 4](https://github.com/advancetoolchain/advance-toolchain/blob/98a33468efe0b253e91dc23945fa7103fdcf9cb0/configs/13.0/packages/gcc/stage_4#L70), looks like it that has been the [default on GCC](https://github.com/gcc-mirror/gcc/blob/7cb40b08bdf739a8e92b9586f5fa765830743522/configure#L8191-L8197) for some time (last change in 2009). So we have been using LTO on stage 3 as well.

If we remove LTO, are there any implications for the packages built with gcc_3? From the logs, these packages are `binutils_2`, `glibc`, `zlib`, `libsphde`, `libdfp` and `gcc_4`

@tuliom pointed to me that this shouldn't impact glibc because it isn't built with LTO. Not sure about the others yet.

**UPDATE**:

From [GCC Compiler Internals](https://gcc.gnu.org/onlinedocs/gccint/LTO-Overview.html):
```
When collect2 detects a link set of .o/.a files with LTO information and the -flto 
is enabled, it invokes lto1 which reads the set of files and aggregates them into
a single translation unit for optimization.
```
Ok, so LTO is only used if our GCC is capable of generating LTO information on object files (`--enable-lto`?) *and* we specify `-flto` when compiling code. We don't use that flag for any of the packages above, so in the end LTO is not used for them.

So disabling LTO in GCC stage 3 shouldn't be a problem after all.

Closes #1732


mscastanho:
Rebased against current master to get the fixes from #1749. Hopefully now all tests will pass.

mscastanho:
Ok, the previous version was applying the `--disable-lto` flag to the symlink for `gcc/stage_3`, which end up changing the file for AT 11.0. With the recent changes to remove GCC stage_2 (#1727) now we have a hard file for `gcc/stage_3`. I changed the PR to correctly apply the flag to this new file instead.

rzinsly (approved):
LGTM

PR #1781, opened 2020-10-28
Fix gdb check on AT <= 12.0

tuliom:
Older AT releases supported both ppc and ppc64.  For those releases,
some executables or symlinks were provided in order to simplify usage.
These executables used to have 2 different prefixes:
 - Target-based, e.g. powerpc64-linux-gnu-readelf;
 - Alternative, e.g. powerpc-linux-gnu-readelf.

However, not all executables available as target-based also had an
alternative-based prefix.  gdb is an exception and is only available as
a target-based prefix.

Futhermore, this issue is not reproduced on AT >= 13.0 because these
versions do not support big-endian targets anymore.

This commit is also improving verbosity of the test showing more
executables that were tested.

Fixes: 2781f58f58d545115ab40bddabcb6b5bbf82ae16
Signed-off-by: Tulio Magno Quites Machado Filho <tuliom@linux.ibm.com>

### DIFF
--- a/fvtr/ck_binaries/ck_binaries.exp
+++ b/fvtr/ck_binaries/ck_binaries.exp
@@ -63,10 +63,10 @@ if { $env(AT_MAJOR_VERSION) >= 9.0 } {
 }
 
 # Binaries distributed with a different name on cross and non-cross
-set dif_exec {cpp elfedit embedspu gdb gprof readelf}
+set dif_exec {cpp elfedit embedspu gprof readelf}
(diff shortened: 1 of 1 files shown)


er-1:
The two failures were because lack of disk space on the build machines. Rebuild in progress.

er-1 (approved):

er-1:
> The two failures were because lack of disk space on the build machines. Rebuilds in progress.

Rebuilds ended well.

PR #1784, opened 2020-10-31
Update gcc on AT next

Cherbot:
Bump to revision 943cc2a1b70f


mscastanho:
Forgot to mention, but I opened a [bug report upstream](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=97709) with the issue and a fix has already been merged. I'm building a new GCC with the fix to check.

mscastanho:
Looks like the problem is fixed. Let me open a new PR with the latest revision.

PR #1795, opened 2020-10-31
Update valgrind on AT 14.0

Cherbot:
Bump to revision 2471e0521da0


er-1:
This would bump the version of Valgrind to 3.17.0.
AT 14.0 should track Valgrind 3.16.*, shouldn't it?

tuliom:
> This would bump the version of Valgrind to 3.17.0.
> AT 14.0 should track Valgrind 3.16.*, shouldn't it?

@er-1 Yes.  Unless we have a good reason to do so.

PR #1797, opened 2020-11-04
Update gcc on AT next

mscastanho:
Bump to revision 1e9f339d946b

Mid-week bump to get fix for ICE from the last tentative update.

Fixes issues seen on #1784


mscastanho:
I'm merging this as it is just a simple version bump as we usually do over the weekend.

PR #1763, opened 2020-10-22
Diverge python on AT next from 14.0

er-1:
python 3.9.0 has been released and will be tracked on AT next.
AT 14.0 will continue to track python 3.8 branch.

Issue #201, opened 2017-10-13
FVTR failure in python (test_ssl) on AT Next on Ubuntu 16.04


gftg85:
Building AT Next on Ubuntu 16.04 causes a FVTR failure in python's test for SSL (test_ssl).

This is the error message:

```
0:01:54 [290/391/1] test_ssl failed -- running: test_multiprocessing_spawn (56 sec), test_multiprocessing_forkserver (63 sec)
Exception in thread Thread-83:
Traceback (most recent call last):
  File "/home/gftg/opt/at-next-12.0-0-alpha/lib64/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/home/gftg/opt/at-next-12.0-0-alpha/lib64/python3.6/test/test_ssl.py", line 1926, in run
    if not self.wrap_conn():
  File "/home/gftg/opt/at-next-12.0-0-alpha/lib64/python3.6/test/test_ssl.py", line 1871, in wrap_conn
    self.sock, server_side=True)
  File "/home/gftg/opt/at-next-12.0-0-alpha/lib64/python3.6/ssl.py", line 401, in wrap_socket
    _context=self, _session=session)
  File "/home/gftg/opt/at-next-12.0-0-alpha/lib64/python3.6/ssl.py", line 808, in __init__
    self.do_handshake()
  File "/home/gftg/opt/at-next-12.0-0-alpha/lib64/python3.6/ssl.py", line 1061, in do_handshake
    self._sslobj.do_handshake()
  File "/home/gftg/opt/at-next-12.0-0-alpha/lib64/python3.6/ssl.py", line 683, in do_handshake
    self._sslobj.do_handshake()
OSError: [Errno 0] Error

test test_ssl failed -- Traceback (most recent call last):
  File "/home/gftg/opt/at-next-12.0-0-alpha/lib64/python3.6/test/test_ssl.py", line 241, in test_random_fork
    self.assertNotEqual(child_random, parent_random)
AssertionError: b')\x8e\xb3[\xd56\xd1\x838\x93P\xdb=\nb>' == b')\x8e\xb3[\xd56\xd1\x838\x93P\xdb=\nb>'
[...]
1 test failed:
    test_ssl
```

This is **not** related to pull request #178.

tuliom:
It seems to be related to [Python issue #28689](https://bugs.python.org/issue28689).

PR #1812, opened 2020-11-13
Follow Python 3.9 on AT next

mscastanho:
This PR includes a fix for Python FVTR tests on AT next, an update to Python 3.9 on AT next and a few changes in the way we distribute Python. Copying the commit message of the most relevant commit:
    
    Python 3.9 has been recently released and is the version that should be
    followed by AT 15 (current AT next). While updating the revision, I took the
    time to fully revise all patches we have been applying to Python and actually
    found out we don't need two of them anymore (with a catch).
    
    1) Patch to skip test_ssl.test_random_fork:
    
        This was needed back in 2017 due to a bug in OpenSSL 1.1.1c (see discussion
        and linked Python issues in [0]). This has been fixed upstream a long time
        ago, and AT next currently follows OpenSSL 1.1.1h, which already includes
        the fix.
    
    2) Patch to update Python's lib path to AT's:
    
        Previous AT versions used to distribute all Python modules and packages
        under a python* directory on /opt/at<XX>/lib64. That is not default behavior
        upstream, so we needed this patch to force AT's Python to look for stuff in
        that subdir instead of the default ones under lib/.
    
        There are differences in the way distros approach this. The default scheme
        of installing everything to /usr/lib/ is used by Debian and Ubuntu. SLES and
        Fedora, on the other hand, split Python stuff into lib and lib64. All
        generic and pure-Python code is installed on lib and any platform-specific
        code goes to lib64. That behavior was not supported by Python upstream's
        install scripts, and these distros had to maintain off-tree patches (like
        ours) for a long time [1]. This changed in Python 3.9, which added the
        --with-platlibdir flag to help Fedora and SLES properly install stuff the
        way they need without needing off-tree patches.
    
        The split into the two different directories is specially useful for
        situations when one needs to install Python built for different targets on
        the same system. This way a 32- and a 64-bit installation can share the
        generic files and only install the platform-specific stuff in their
        corresponding directories. That obviously is not the case for AT, as we
        don't support 32-bit targets anymore.
    
        Still, I'd prefer if AT used an installation method that is supported
        upstream, so we can reduce the amount of off-tree patches we need to
        maintain, specially a big and complex one like this. They usually need
        updates now and then when a new version makes incompatible changes.
        Splitting the installation into lib and lib64 dirs as described above makes
        sense and should work well for AT.
    
        So instead of applying our off-tree patch, this commit configures Python
        with --with-platlibdir=lib64, which as a side-effect changes the way AT
        distributes Python modules and packages:
          Before: everything under /opt/at<XX>/lib64/
          After: generic and pure-Python modules /opt/at<XX>/lib and platform
          specific files in /opt/at<XX>/lib64.
    
        This change should be transparent to AT users as Python configured this way
        knows where it should look for stuff.
    
    [0] https://github.com/advancetoolchain/advance-toolchain/issues/201
    [1] https://bugs.python.org/issue1294959

Example of how Python stuff is distributed in AT 14:
``` bash
$ find /opt/at14.0/ -name '*python*' -type d
/opt/at14.0/lib/debug/opt/at14.0/lib64/python3.8
/opt/at14.0/lib64/python3.8
/opt/at14.0/lib64/cmake/boost_python-1.73.0
/opt/at14.0/share/gcc-10.2.1/python
/opt/at14.0/share/gdb/python
/opt/at14.0/include/python3.8
/opt/at14.0/include/python3.8/cpython
/opt/at14.0/include/boost/mpi/python
/opt/at14.0/include/boost/parameter/aux_/python
/opt/at14.0/include/boost/python
```

Now on AT next after this change:
```
$ find ./opt/at-next-15.0-0-alpha/ -name '*python*' -type d
./opt/at-next-15.0-0-alpha/lib64/cmake/boost_python-1.73.0
./opt/at-next-15.0-0-alpha/lib64/python3.9
./opt/at-next-15.0-0-alpha/include/python3.9
./opt/at-next-15.0-0-alpha/include/python3.9/cpython
./opt/at-next-15.0-0-alpha/include/boost/mpi/python
./opt/at-next-15.0-0-alpha/include/boost/parameter/aux_/python
./opt/at-next-15.0-0-alpha/include/boost/python
./opt/at-next-15.0-0-alpha/share/gcc-11.0.0/python
./opt/at-next-15.0-0-alpha/share/gdb/python
./opt/at-next-15.0-0-alpha/lib/python3.9
```

Closes #1763 
Closes #1818

### DIFF
--- a/fvtr/gdb/gdb.exp
+++ b/fvtr/gdb/gdb.exp
@@ -27,7 +27,7 @@ if { [array names env -exact "AT_GDB_VER"] == "" } {
 # AT 5.0 and 6.0 used to use development versions of GDB, instead of stable
 # versions like AT >= 7.0.  Which causes issues when trying to run this
 # testcase, like when testing for GDB version.
-if { [regexp "5\..*|6\..*" $env(AT_MAJOR_VERSION)] } {
+if { [regexp "\A5\..*|\A6\..*" $env(AT_MAJOR_VERSION)] } {
 	printit "This test doesn't work on AT 5.0 or 6.0." $WARNING
 	printit "Skipping..."
 	exit $ENOSYS
--- a/configs/next/packages/python/sources
+++ b/configs/next/packages/python/sources
@@ -20,8 +20,8 @@
 #
 
 ATSRC_PACKAGE_NAME="Python"
-ATSRC_PACKAGE_VER=3.8.6
-ATSRC_PACKAGE_REV=a12f459ec2a3
+ATSRC_PACKAGE_VER=3.9.0
+ATSRC_PACKAGE_REV=7ae19ef5cf5d
 ATSRC_PACKAGE_LICENSE="Python Software Foundation License 2"
 ATSRC_PACKAGE_DOCLINK="https://docs.python.org/${ATSRC_PACKAGE_VER%\.*}/"
 ATSRC_PACKAGE_RELFIXES=
@@ -43,21 +43,10 @@ ATSRC_PACKAGE_BUNDLE=toolchain_extra
 
 atsrc_get_patches ()
 {
-    # Update python lib path to AT's
-	at_get_patch \
-		https://raw.githubusercontent.com/powertechpreview/powertechpreview/cd1a9a7bb398332641b05a45e12807ce6dabcc19/Python%20Fixes/python-3.8-getlib64.patch \
-		a8c6cab1ed2d9cabdb7369ae4b04a16a || return ${?}
-
-	# Disable test_random_fork (test_ssl).
-	# See https://github.com/advancetoolchain/advance-toolchain/issues/201
-	at_get_patch \
-		https://raw.githubusercontent.com/powertechpreview/powertechpreview/89d1d3cb24e43f8ec716e1047f8d94702b131fda/Python%20Fixes/python-3.6.3-skip_test_random_fork.patch \
-		b6713b5bea7d419b18b8f8319ba5a6f2 || return ${?}
-
 	# Increase a timeout in test_subprocess.
 	at_get_patch \
-		https://raw.githubusercontent.com/powertechpreview/powertechpreview/ec0f7202f72ca7b946525b29aca58ec50979906e/Python%20Fixes/python-3.6.12-timeout-test_subprocess.patch \
-		a50e89dca0c51249be26f027a0474a23 || return ${?}
+	        https://raw.githubusercontent.com/powertechpreview/powertechpreview/b9f88b42e734110e69873ea810272facc85b97c0/Python%20Fixes/python-3.9-timeout-test_subprocess.patch \
+	        92cfc34136fafb9420655b4bd3ef39d1 || return ${?}
 
 	# Disable zlib version check (test_zlib).
 	at_get_patch \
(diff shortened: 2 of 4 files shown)


tuliom:
I miss the directory with the debug information:
```
/opt/at14.0/lib/debug/opt/at14.0/lib64/python3.8
```
Did it really disappear or is just just the difference between installing via RPM and building manually?

mscastanho:
@tuliom I think the debug dir is a symlink, and it's only created during the install step.
```
$ find ./opt/at-next-15.0-0-alpha/ -name 'debug' -type d
./opt/at-next-15.0-0-alpha/lib64/go/11.0.0/powerpc64le-linux-gnu/debug
./opt/at-next-15.0-0-alpha/include/c++/11.0.0/debug
./opt/at-next-15.0-0-alpha/include/boost/bimap/detail/debug
./opt/at-next-15.0-0-alpha/include/boost/preprocessor/debug
./opt/at-next-15.0-0-alpha/include/boost/spirit/home/classic/debug
```

mscastanho:
Update to also apply the FVTR test fix to gdb, which suffered from the same issue.

tuliom:
> Did it really disappear or is just just the difference between installing via RPM and building manually?

Confirmed! This is just a different between comparing an installed AT vs. a directory created at build time.
So, the debug directory was not affected by this change.
Files are under `opt/at-next-15.0-0-alpha/lib64/python3.9`.

tuliom (approved):
LGTM.
Let's just wait for all the tests to pass before merging it.

fvtr/gdb/gdb.exp
```diff
@@ -27,7 +27,7 @@ if { [array names env -exact "AT_GDB_VER"] == "" } {
 # AT 5.0 and 6.0 used to use development versions of GDB, instead of stable
 # versions like AT >= 7.0.  Which causes issues when trying to run this
 # testcase, like when testing for GDB version.
-if { [regexp "5\..*|6\..*" $env(AT_MAJOR_VERSION)] } {
+if { [regexp "\A5\..*|\A6\..*" $env(AT_MAJOR_VERSION)] } {
```
ThinkOpenly, line 30: Completely subjective and perhaps personal preference, but would `^` work here instead of `\A`? The caret (`^`) is more familiar to me, at least.
(Same for the next change, as well.)

ThinkOpenly (approved):
For the commit "Fix Python and gdb FVTR tests on AT next", might I suggest a new subject "Fix to not exclude Python and gdb FVTR tests", since you are not fixing the tests themselves. Not a big deal, though.

fvtr/gdb/gdb.exp
mscastanho: @ThinkOpenly It does, but apparently requires an extra switch (`-lineanchor`). But I agree that is more standard and can make it less confusing to future us/others taking a look at this code.

But since tests are already running (and they take a long time), I'll leave that to a future PR addressing #1813.

mscastanho:
The `test_subprocess` Python test failed on RHEL 8. I'm taking a closer look to understand what happened.

tuliom:
> The `test_subprocess` Python test failed on RHEL 8. I'm taking a closer look to understand what happened.

@mscastanho Reminder: that's the test that used to timeout.

mscastanho:
@tuliom Yes, I'm aware. It timed out again, but on a different part. I'm increasing the timeout and testing it again.

mscastanho:
I updated the timeout for another Python test, rebased the PR against master and also updated the commit title as suggested by @ThinkOpenly.

mscastanho:
Info about failed tests:
- AT 14.0 RHEL8 ppc64le: server under high load and some processes got killed. Trying again.
- AT * SLES12 ppc64le: server was misconfigured and `oprofile` test failed on all of them. Fixed and trying again.
- AT 11 Ubuntu16 ppc64le: intermittent Python test failures.
- AT 12 Debian9 ppc64le: same...

mscastanho:
Last failing tests:
- AT 12 Debian9: Rebuilding to see if we avoid the intermittent failures this time
- AT 10 SLES12: One python test is consistently failing: `test_strptime`. This test has been fixed on Python 3.6 (AT >= 11) by python/cpython@f1b9ad3d38c11. That specific build server has been update recently and probably altered some timezone files that triggered the issue. I'll include a commit to skip this test on AT 10 on an upcoming PR.

mscastanho:
AT 12 Debian9 build passed.

PR #1815, opened 2020-11-14
Update gcc on AT next

Cherbot:
Bump to revision dcf0dde488b8

Issue #1691, opened 2020-09-16
Remove -Wno-error flags from binutils stage_2 on AT next


mscastanho:
`-Wno-error flags` has been added by PR #1690 (still pending) to avoid an error when building gold, caused by the deprecation of `mallocinfo` on glibc.

See #1659 and [BZ#26585](https://sourceware.org/bugzilla/show_bug.cgi?id=26585) for more info.

We should be able to remove the flag once the issue has been fixed upstream.

All -Wno-error flags are expected to be removed, including `-Wno-error flags` and `-Wno-error=stringop-overflow`.

PR #1823, opened 2020-11-16
next: Update GCC and fix binutils build

mscastanho:
With the latest bump, GCC started throwing the following warning when building binutils stage 2:
    
    ./gold/plugin.cc:528:10: error: ‘char* strncpy(char*, const char*, size_t)’
    specified bound depends on the length of the source argument [-Werror=stringop-overflow=]
      528 |   strncpy(tempdir, dir_template, len);
          |   ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~
    ./gold/plugin.cc:526:22: note: length computed here
      526 |   size_t len = strlen(dir_template) + 1;
          |                ~~~~~~^~~~~~~~~~~~~~
    
The code causing the error is not necessarily wrong, we can ignore this
warning for the time being.

Closes #1815


tuliom (approved):
LGTM.
I updated issue #1691 so that we remove this flag later too.

It's best to wait for tests to complete before merging this, though.

mscastanho:
All tests passed. Merging...

PR #1828, opened 2020-11-21
Update gdb on AT next

Cherbot:
Bump to revision abd20cb63700


mscastanho:
The failing builds were not able to find a proper GMP library during the configure step. GMP became a requirement to build gdb in this last set of commits we are trying to merge. A new build flag `--with-libgmp-prefix` was added to allow specifying the dir where gdb can find GMP. I'm preparing a PR to add that flag and point to AT's GMP.

PR #1874, opened 2020-12-14
fvtr: Increase timeout for Python test_eintr test

### LINKED ISSUE #1873: test_eintr Python test failing intermittently
`test_eintr` has been failing intermittently on our FVTR test suite, like on #1866. On most cases, re-running the build was enough to get the test to pass, so this is likely another timeout-related issue. This needs further investigation.

mscastanho:
This test has been failing intermittently on our FVTR Python tests when the
server is under heavy load (like when testing multiple PRs during the
weekend). So extend the tolerance to avoid these spurious failures.

(Hopefully)
Fixes #1873

### DIFF
--- a/configs/11.0/packages/python/sources
+++ b/configs/11.0/packages/python/sources
@@ -62,6 +62,11 @@ atsrc_get_patches ()
 		https://raw.githubusercontent.com/powertechpreview/powertechpreview/1fa3a5896a6ce15e8b85fcf24109db5c1ea5548a/Python%20Fixes/python-3.7-disable-zlib-version-check.patch \
 		f52555a37aec6cb7f43eb052990e051b || return ${?}
 
+	# Increase a timeout in test_eintr
+	at_get_patch \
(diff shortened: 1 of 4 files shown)


er-1 (approved):

mscastanho:
The test failed again. I'm not sure if just increasing the timeout will be enough. If it takes over 2 minutes to complete this simple test, there must be more going on...

tuliom:
@mscastanho I had prepared the exact same patch and had the exact same conclusions: there must be something else going on.
I reviewed the settings of the VMs and fixed an issue that could be causing this, but it's still happening.
I'm afraid the issue is that we're starting too many builds at the same time on a single host.
I think we may need to:
1. Stop creating more VMs on hosts that are already under high load.
2. Kill old VMs on that host.
3. Reduce the number of jobs on some VMs.
4. Migrate VMs to another host.
5. If we can't kill or move too many VMs, we may need to add more hosts.

PR #1852, opened 2020-12-05
Update gcc on AT next

Cherbot:
Bump to revision 625e002396f7

PR #1875, opened 2020-12-16
Update gcc with proper required fixes

mscastanho:
Now that the issues seen on #1852 have been fixed upstream, I can finally update the GCC revision. However, this last set of gcc commits introduced some change in the preprocessor behaviour that is causing glibc build failures. For that reason, I also had to add a `-Wno-error=stringop-overread` to glibc configs. More info on #1876.

Finally, I had to regenerate the libssp patch to properly apply to the new GCC revision.

Closes #1852


mscastanho:
Updated to a newer GCC revision, as the previous one was affected by `libcody` issues that appeared upstream this week. They should be fixed now.

mscastanho:
Turns out we only needed to ignore the warning on glibc stage 1 for cross builds, but on all stages for natives builds. Updated and pushed changes.

er-1 (approved):

PR #1900, opened 2021-01-09
Update gcc on AT 14.0

Cherbot:
Bump to revision 2093e873bb6c


mscastanho:
Again, libssp patch needs an update

PR #1903, opened 2021-01-12
Fix update_revision.sh to get updates from HEAD

### LINKED ISSUE #1897: Cherbot stopped updating some packages from AT next
The last update is from Dec. 5th : #1852 .
This is affecting all AT releases.

er-1:
After commit 8607a59, the script wasn't able to get updates from HEAD branch
of packages because it tried to get refs/heads/HEAD instead of HEAD.

Fix #1897

### DIFF
--- a/scripts/utilities/update_revision.sh
+++ b/scripts/utilities/update_revision.sh
@@ -158,8 +158,6 @@ get_latest_revision ()
 		# relocated within the vendors directory.
 		if [[ $package == "gcc" ]] && [[ $configset != "next" ]]; then
 		    branch="refs/vendors/ibm/heads/${branch}"
-		else
-		    branch="refs/heads/${branch}"
 		fi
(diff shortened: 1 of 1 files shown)


scripts/utilities/update_revision.sh
```diff
@@ -158,8 +158,6 @@ get_latest_revision ()
 		# relocated within the vendors directory.
 		if [[ $package == "gcc" ]] && [[ $configset != "next" ]]; then
 		    branch="refs/vendors/ibm/heads/${branch}"
-		else
```
mscastanho, line 161: Ok. This `else` should not be here because it is overwriting the value we have set above when `ATSRC_PACKAGE_BRANCH` is not set.

mscastanho (approved):
LGTM

tuliom (approved):

PR #1908, opened 2021-01-16
Update gcc on AT next

Cherbot:
Bump to revision 2c356f221bba

PR #1924, opened 2021-01-21
Temporarily disable boost on AT next + Update GCC

mscastanho:
A recent GCC update is causing boost build to fail. The fix may take a
few days, so meanwhile let's stop building boost so our builds are able
to finish and we can keep bumping the GCC version.

Closes #1908


tuliom (approved):
LGTM.
I think it's a good idea to create an issue in order to not forget to re-enable this later.

mscastanho:
@tuliom Agreed. #1925 created for that matter.

PR #1937, opened 2021-01-27
Add a script to check upstream version

er-1:
New script that compares the upstream version number with the version number
stored in the given sources file.

### DIFF
--- a/scripts/utilities/check_versions.sh
+++ b/scripts/utilities/check_versions.sh
@@ -1,0 +1,108 @@
+#!/bin/bash
+#
+# Copyright 2021 IBM Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


mscastanho (approved):
LGTM.
Should we also add zlib to the list of packages? I know it hasn't seen releases for *a while*, but we hope it will in the future.

er-1:
> Should we also add zlib to the list of packages? I know it hasn't seen releases for _a while_, but we hope it will in the future.

Yes. It'd be a good thing. May be also take advantage to update https://github.com/advancetoolchain/advance-toolchain/blob/master/configs/next/packages/zlib/sources (git clone instead of wget...).

tuliom (commented):
I noticed that boost, gmp, libnxz, sphde, libvecpf, mpc, mpfr, paflib, pveclib are also missing.

I don't think they're required to be available in this patch, but it would be nice to have a rationale on why they were left behind even if it's just: they will be added later.

er-1:
> I noticed that boost, gmp, libnxz, sphde, libvecpf, mpc, mpfr, paflib, pveclib are also missing.

About gmp, mpfr and mpc, we agreed to not update them (commit 7f628a3b9e7f14b150a84f1f7e1c59939372f13c) to keep versions close to what gcc requires.

That said, I agree that seems awkward some packages are missing.They will be added later.

PR #1953, opened 2021-02-10
Handle new curl formatting for status codes on package update script

mscastanho:
Apparently curl has changed the way it returns the status code, from
adding a header-like line with 'Status: $code $description' to a
single line at the top of the output like 'HTTP/$version $code'.

This commit makes the necessary adjustments to update_revision.sh to
handle this transition, and should work with older versions of curl as
well.

#1952 was created using the updated script.

### DIFF
--- a/scripts/utilities/update_revision.sh
+++ b/scripts/utilities/update_revision.sh
@@ -117,6 +117,28 @@ print_msg ()
 	echo ${2}
 }
 
+# This function makes a request to the GitHub REST API. It returns the HTTP
+# status code for the request and saves any output to outfile. Any
+# extra parameters are passed directly as arguments to curl.
+#
+# Parameters:
+#     $1 - outfile
+#     $2 - endpoint
+ghapi_call ()
+{
+	if [[ ${#} -lt 2 ]]; then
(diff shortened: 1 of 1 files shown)


er-1 (approved):

scripts/utilities/update_revision.sh
```diff
@@ -294,6 +316,9 @@ send_to_github ()
 	pkg=$(echo ${1} | awk -F "/" '{ print $(NF-1) }')
 	cfg=$(echo ${1} | awk -F "/" '{ print $(NF-3) }')
 
+        # Temp file to save output of REST calls to the GitHub API
+	out=$(mktemp '/tmp/ghapi-XXXXX.out')
```
tuliom, line 320: Isn't it important to remove this file at the end of the execution?

tuliom (commented):
Looks good to me, but I have a question.

scripts/utilities/update_revision.sh
mscastanho: I added a `trap` to guarantee it gets deleted.

tuliom (approved):

PR #1977, opened 2021-02-27
Update gcc on AT next

Cherbot:
Bump to revision 06a9f20f97e7

PR #1979, opened 2021-02-27
Update glibc on AT next

Cherbot:
Bump to revision 035c012e32c1


mscastanho:
Failed when running the configure step for GCC stage 3:
```
configure:4613: /home/tcbot/bot-slave/at_opt/at-next-15.0-0-alpha/bin/gcc -o conftest -O2 -mminimal-toc  -L/home/tcbot/bot-slave/at_opt/at-next-15.0-0-alpha/tmp/gcc_3 conftest.c  >&5
configure:4617: $? = 0
configure:4624: ./conftest
./conftest: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by ./conftest)
configure:4628: $? = 1
configure:4635: error: in `/home/tcbot/bot-worker/at/src/at15.0-0-alpha.redhat-8_ppc64le_ppc64le/builds/gcc_3':
configure:4637: error: cannot run C compiled programs.
```
Looks like this test program tried to link against the system's glibc and didn't find a glibc 2.34 symbol. But why is it doing that? The loader should have used AT's glibc.

mscastanho:
A summary of what is happening when the error happens. The contents of `conftest.c` come from GCC's `configure` file:
```
[mscastanho@gauss gcc_3]$ cat conftest.c
#include <stdio.h>
int
main ()
{
FILE *f = fopen ("conftest.out", "w");
 return ferror (f) || fclose (f) != 0;

  ;
  return 0;
}

[mscastanho@gauss gcc_3]$ /home/mscastanho/AT/broken-gcc-config/opt/at-next-15.0-0-alpha/bin/gcc -o conftest -O2 -mminimal-toc  -L/home/mscastanho/AT/broken-gcc-config/opt/at-next-15.0-0-alpha/tmp/gcc_3 conftest.c
[mscastanho@gauss gcc_3]$ ./conftest 
./conftest: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by ./conftest)
[mscastanho@gauss gcc_3]$ readelf -Ws ./conftest | grep 'GLIBC_2.34'
     1: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND __libc_start_main@GLIBC_2.34 (2)     [<localentry>: 8] 
    19: 00000000100004e0     0 NOTYPE  LOCAL  DEFAULT   12 00000042.plt_call.__libc_start_main@@GLIBC_2.34
    26: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND __libc_start_main@GLIBC_2.34         [<localentry>: 8]
[mscastanho@gauss gcc_3]$ file ./conftest
./conftest: ELF 64-bit LSB executable, 64-bit PowerPC or cisco 7500, version 1 (SYSV), dynamically linked, interpreter /lib64/ld64.so.2, for GNU/Linux 4.18.0, with debug_info, not stripped
```
The script uses AT's GCC from stage 1, but for some reason the executable generated is using the system's loader, which ends up linking the code against the system's glibc. [A recent glibc commit](https://sourceware.org/git/?p=glibc.git;a=commit;h=035c012e32c11e84d64905efaf55e74f704d3668) added the symbol `__libc_start_main@@GLIBC_2.34`, which is being required by the executable, but is not found in the system's (older) glibc.

So the code is being compiled against AT's glibc, but loaded by and dynamically linked against the system's glibc.

Are we missing some configuration on GCC stage 1 to properly set the `PT_INTERP` section on generated executables to guarantee they use AT's loader?

mscastanho:
> Are we missing some configuration on GCC stage 1 to properly set the PT_INTERP section on generated executables to guarantee they use AT's loader?

Setting `PT_INTERP` to AT's loader is part of what `--with-advance-toolchain` GCC configure flag does.

https://github.com/gcc-mirror/gcc/blob/4ed0a92f6cfc647e2ad8ceaa1e5709545c915465/gcc/config.gcc#L5092-L5093

That flag is used for GCC stages 3 and 4 on AT, but not for stage 1.

tuliom:
Does GCC allow to set the interpreter when running the configure? e.g. via `--with-stage1-ldflags=`

Source: https://gcc.gnu.org/install/configure.html

mscastanho:
`--with-stage1-ldflags=` is already being used there:
https://github.com/advancetoolchain/advance-toolchain/blob/60c3f27e348d7fa0c90334898c55171aa6765ffc/configs/next/packages/gcc/stage_3#L145

So that's not enough. It's not used on that specific point on configure. I noticed it uses `CFLAGS` to compile the test program. I'm testing passing the proper `-Wl,--dynamic-linker=` there.

mscastanho:
Forcing the dynamic linker to use on gcc_3 works past the problem. This is the change I made:
```diff
diff --git a/configs/next/packages/gcc/stage_3 b/configs/next/packages/gcc/stage_3
old mode 100644
new mode 100755
index ab59a31..5ad710e
--- a/configs/next/packages/gcc/stage_3
+++ b/configs/next/packages/gcc/stage_3
@@ -106,8 +106,8 @@ atcfg_configure() {
                AR="${at_dest}/bin/ar" \
                NM="${at_dest}/bin/nm" \
                RANLIB="${at_dest}/bin/ranlib" \
-               CFLAGS="-O2 -mminimal-toc" \
-               CXXFLAGS="-O2 -mminimal-toc" \
+               CFLAGS="-O2 -mminimal-toc -Wl,--dynamic-linker=${ldso}" \
+               CXXFLAGS="-O2 -mminimal-toc -Wl,--dynamic-linker=${ldso}" \
                LDFLAGS="-L${gcclibdir}" \
                POSTSTAGE1_LDFLAGS="-L${gcclibdir} -lstdc++ -lsupc++ -lm -lgmp -lgmpxx -lmpfr -lmpc" \
                ${ATSRC_PACKAGE_WORK}/configure \
```
But now gcc_4 fails when building libgo:
```
/home/mscastanho/AT/glibc-breaks-gcc3/build/at15.0-0-alpha.redhat-8_ppc64le_ppc64le/sources/gcc/libgo/runtime/proc.c: In function ‘runtime_malg’:
/home/mscastanho/AT/glibc-breaks-gcc3/build/at15.0-0-alpha.redhat-8_ppc64le_ppc64le/sources/gcc/libgo/runtime/proc.c:805:38: error: comparison of integer expressions of different signedness: ‘uintptr’ {aka ‘long unsigned int’} and ‘long int’ [-Werror=sign-compare]
  805 |                         if(stacksize < SIGSTKSZ)
      | 
```
I think https://github.com/gcc-mirror/gcc/commit/331763de7d4850702a0f67298f36017c73cdb103 could be the fix for it, but I'm afraid that when we update to this gcc revision, we may hit the GCC issue from #1977 and #1990. Testing now.

PR #1990, opened 2021-03-02
Update gcc on AT next

er-1:
Bump to version 11.0.1
Bump to revision ~~06a9f20f97e7~~  f1d012911bd1

Close #2020 


mscastanho:
@er-1 Thanks for fixing the version string. Now we just need to wait for the fix for the cross compiler upstream so we can use that new revision. A [v2](https://gcc.gnu.org/pipermail/gcc-patches/2021-March/566023.html) is already under review

er-1:
Failure are related with #2017

mscastanho:
@er-1 Could you update the revision to `f1d012911bd103e9701ed378f32928a3cbcf028a`? It should contain the fix [1] and pass all tests.

https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=f1d012911bd103e9701ed378f32928a3cbcf028a

er-1:
> @er-1 Could you update the revision to `f1d012911bd103e9701ed378f32928a3cbcf028a`? It should contain the fix [1] and pass all tests.

Done.

mscastanho (approved):

PR #1991, opened 2021-03-04
Fix next to work without RUNPATH on binaries

lucmaga:
This is still a draft

I still need to test this on RHEL and SUSE yet. Works on Debian.

For this to work we need to apply the patch bellow on gcc.

```
diff --git a/gcc/config.gcc b/gcc/config.gcc
index 17fea83b2e4..6827b799dc6 100644
--- a/gcc/config.gcc
+++ b/gcc/config.gcc
@@ -5071,16 +5071,6 @@ case "${target}" in
                        (at="/opt/$with_advance_toolchain"
                         echo "/* Use Advance Toolchain $at */"
                         echo
-                        echo "#undef  LINK_OS_EXTRA_SPEC32"
-                        echo "#define LINK_OS_EXTRA_SPEC32" \
-                             "\"%(link_os_new_dtags)" \
-                             "-rpath $prefix/lib -rpath $at/lib\""
-                        echo
-                        echo "#undef  LINK_OS_EXTRA_SPEC64"
-                        echo "#define LINK_OS_EXTRA_SPEC64" \
-                             "\"%(link_os_new_dtags)" \
-                             "-rpath $prefix/lib64 -rpath $at/lib64\""
-                        echo
                         echo "#undef  LINK_OS_NEW_DTAGS_SPEC"
                         echo "#define LINK_OS_NEW_DTAGS_SPEC" \
                              "\"--enable-new-dtags\""
```

### DIFF
--- a/scripts/distributed/find_dependencies.sh
+++ b/scripts/distributed/find_dependencies.sh
@@ -82,13 +82,13 @@ for file in "${@}"; do
 		exit 1
 	fi;
 
-	# Get the RUNPATH.  The RUNPATH will tell us if this file was built
-	# using AT or not.
-	runpath=$(${readelf_path} -d ${file} \
-			| grep 'RUNPATH' \
-			| sed 's/^.*\: \[\(.*\)\]$/\1/')
-	at_path=${runpath%%/lib*}
-	if [[ ${is_cross} != "yes" && ${runpath} == /opt/at* ]]; then
+	# Get the interpreter.  The interpreter will tell us if this file was
+	# built using AT or not.
+	interpreter=$(${readelf_path} -e ${file} \
+			| grep 'interpreter' \
+			| sed 's/\[.*\: \(.*\)\]/\1/')
+	at_path=${interpreter%%/lib*}
+	if [[ ${is_cross} != "yes" && ${interpreter} == /opt/at* ]]; then
 		ldd_path=${at_path}/bin/ldd
 	else
 		ldd_path=$(which ldd)
--- a/configs/next/deb/monolithic/rules
+++ b/configs/next/deb/monolithic/rules
@@ -1,1 +1,88 @@
-../../../14.0/deb/monolithic/rules
+#!/usr/bin/make -f
+#
+# Copyright 2017 IBM Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+PACKAGES=$(shell dh_listpackages)
+
+build:
+	dh_testdir
+	dh_testroot
+	dh_prep
+
+clean:
+	dh_testdir
+	dh_testroot
+	dh_clean -d
+
+binary-indep: build
+
+binary-arch: build binary-cp
+	dh_installdirs
+	dh_installdocs
+	dh_installchangelogs
+ifeq (__USE_SYSTEMD__, yes)
+	dh_systemd_enable
+endif
+
+# Copy the packages' files.
+	dh_install
+	dh_missing --fail-missing
+ifeq (__USE_SYSTEMD__, yes)
+	dh_systemd_start
+endif
+
+# Strip binaries
+	perl debian/dh_strip --dest=__AT_DEST__ --exclude=getconf --exclude=libgo. --exclude=ld- --dbg-package=advance-toolchain-__AT_MAJOR_INTERNAL__-runtime-dbg -p advance-toolchain-__AT_MAJOR_INTERNAL__-runtime
+	perl debian/dh_strip --dest=__AT_DEST__ --dbg-package=advance-toolchain-__AT_MAJOR_INTERNAL__-mcore-libs-dbg -p advance-toolchain-__AT_MAJOR_INTERNAL__-mcore-libs
+	perl debian/dh_strip --dest=__AT_DEST__ --exclude=getconf --dbg-package=advance-toolchain-__AT_MAJOR_INTERNAL__-devel-dbg -p advance-toolchain-__AT_MAJOR_INTERNAL__-devel
+	perl debian/dh_strip --dest=__AT_DEST__ --dbg-package=advance-toolchain-__AT_MAJOR_INTERNAL__-perf-dbg -p advance-toolchain-__AT_MAJOR_INTERNAL__-perf
+	perl debian/dh_strip --dest=__AT_DEST__ --dbg-package=advance-toolchain-__AT_MAJOR_INTERNAL__-libnxz-dbg -p advance-toolchain-__AT_MAJOR_INTERNAL__-libnxz
+
+	dh_compress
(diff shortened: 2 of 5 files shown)


mscastanho:
@lamm Some suggestions:
- Commit with `-s` to have the signed off by line on the commit message (required by DCO check)
- When breaking a symlink + altering the new file, do it in separate commits, it makes it easier to review. Example: 186e99a4cc
- You should modify `configs/next/package/gcc/sources` to apply that patch to GCC. As I see now the PR is not functional as it is
- If you do the last step, the CI will be able to properly test your PR (in all configs affected) when you do your next forced push.

scripts/distributed/find_dependencies.sh
```diff
@@ -85,8 +85,8 @@ for file in "${@}"; do
 	# Get the RUNPATH.  The RUNPATH will tell us if this file was built
```
tuliom, line 85: Need to update this comment too.

tuliom (changes requested):
I like this proposal.
Matheus already raised some important points.
Besides them and the fact that RHEL and SLES need to be tested too, I think it's important to test the same thing for our cross compilers.

lucmaga:
>     * You should modify `configs/next/package/gcc/sources` to apply that patch to GCC. As I see now the PR is not functional as it is

I can do this for the CI. However I still don't know if it's best to send the GCC patch upstream first or use it downstream for now.

lucmaga:
Update: This patchset is working on RHEL and SLES as well.

tuliom:
> I can do this for the CI. However I still don't know if it's best to send the GCC patch upstream first or use it downstream for now.

@mscastanho I suspect there is something else preventing this PR from being tested.
AFAICS, this PR is fully contained and correct regardless if GCC is modified or not.
However, if GCC is modified, we might need other changes, e.g. remove the RUNPATH patch from glibc.

I do expect GCC and glibc to be modified, but I don't think that has to happen on the same PR.

lucmaga:
@tuliom @mscastanho This is marked as a draft. I guess that's why it's not being tested.

mscastanho:
@tuliom @lucmaga there is an issue with our CI that is preventing it from testing PRs. I'm testing a fix right now...

scripts/distributed/find_dependencies.sh
```diff
@@ -82,11 +82,11 @@ for file in "${@}"; do
...
-	# Get the RUNPATH.  The RUNPATH will tell us if this file was built
-	# using AT or not.
-	runpath=$(${readelf_path} -d ${file} \
-			| grep 'RUNPATH' \
-			| sed 's/^.*\: \[\(.*\)\]$/\1/')
+	# Get the interpreter.  The interpreter will tell us if this file was
+	# built using AT or not.
+	runpath=$(${readelf_path} -e ${file} \
+			| grep 'interpreter' \
+			| sed 's/\[.*\: \(.*\)\]/\1/')
```
mscastanho, line 89: I tested the code before the last push, when it still used `readelf -d`, which did not show any `interpreter` in the output. I'd expect that to generate an RPM file with some wrong dependencies, but comparing the rpms and spec files between this and a recent build I had for #2008, looks like they were exactly the same.

Comparing the dependencies in the RPM files:
```
$ lammrpms=AT/valgrind-libexec/build/at15.0-0-alpha.redhat-8_ppc64le_ppc64le/rpms/ppc64le
$ valrpms=AT/valgrind-libexec/build/at15.0-0-alpha.redhat-8_ppc64le_ppc64le/rpms/ppc64le
$ for rpm in $(ls $lammrpms); do diff -s <(rpm -qp $lammrpms/${rpm}) <(rpm -qp $valrpms/${rpm}) ; done
Files /dev/fd/63 and /dev/fd/62 are identical
Files /dev/fd/63 and /dev/fd/62 are identical
Files /dev/fd/63 and /dev/fd/62 are identical
Files /dev/fd/63 and /dev/fd/62 are identical
Files /dev/fd/63 and /dev/fd/62 are identical
Files /dev/fd/63 and /dev/fd/62 are identical
Files /dev/fd/63 and /dev/fd/62 are identical
Files /dev/fd/63 and /dev/fd/62 are identical
Files /dev/fd/63 and /dev/fd/62 are identical
Files /dev/fd/63 and /dev/fd/62 are identical
Files /dev/fd/63 and /dev/fd/62 are identical
Files /dev/fd/63 and /dev/fd/62 are identical
```
(notice I used temporary redirection files to compare the outputs, that's why the names with fds).

Comparing the spec files:
```
$ diff -s AT/{lamm-runpath-pr,valgrind-libexec}/build/tmp.at15.0-0-alpha.redhat-8_ppc64le_ppc64le/rpmbuild/SPECS/advance-toolchain.spec 
Files AT/lamm-runpath-pr/build/tmp.at15.0-0-alpha.redhat-8_ppc64le_ppc64le/rpmbuild/SPECS/advance-toolchain.spec and AT/valgrind-libexec/build/tmp.at15.0-0-alpha.redhat-8_ppc64le_ppc64le/rpmbuild/SPECS/advance-toolchain.spec are identical
$ diff -s AT/{lamm-runpath-pr,valgrind-libexec}/build/tmp.at15.0-0-alpha.redhat-8_ppc64le_ppc64le/rpmbuild/SPECS/advance-toolchain_at-compat.spec
79,80c79,80
< mkdir -p ${RPM_BUILD_ROOT}/home/mscastanho/AT/lamm-runpath-pr/opt/at14.0/lib \
<          ${RPM_BUILD_ROOT}/home/mscastanho/AT/lamm-runpath-pr/opt/at14.0/lib64
---
> mkdir -p ${RPM_BUILD_ROOT}/home/mscastanho/AT/valgrind-libexec/opt/at14.0/lib \
>          ${RPM_BUILD_ROOT}/home/mscastanho/AT/valgrind-libexec/opt/at14.0/lib64
102,103c102,103
< /home/mscastanho/AT/lamm-runpath-pr/opt/at14.0/lib
< /home/mscastanho/AT/lamm-runpath-pr/opt/at14.0/lib64
---
> /home/mscastanho/AT/valgrind-libexec/opt/at14.0/lib
> /home/mscastanho/AT/valgrind-libexec/opt/at14.0/lib64
```
(the only difference here is the destination path, which is expected).

So I guess this is working?

But I suggest renaming the `runpath` variable to `interpreter` on lines 87, 90 and 91. It can cause confusion in the future since runpath is not used anymore.

mscastanho (changes requested):
I like the proposal in this PR, and the code overall looks good. I just have some suggestions.

As I understand it, this PR does not remove the usage of RUNPATH from AT. It just adapts the code to work when that patch gets merged upstream and integrated into AT. I'm fine with that, but it should be clear in the commit message that this patch is a preparation for that. A (brief) context about this change (in the commit message) would also be helpful for future reference.

Also, the commit message states that AT doesn't use `RUNPATH` anymore (`...it uses the RUNPATH to find out the libraries. As we don't have it anymore...`), which is not true (yet), so that should be reworded.

lucmaga:
After some discussions we decided to add the GCC patch on AT next for some time before sending to upstream.
I just updated the PR with the inclusion of the necessary changes.

tuliom:
Both tests `AT 12.0 Debian9 ppc64le 341` and `AT 12.0 RHEL7 ppc64le 506` failed because of intermittent python failures.
I'm restarting the tests in order to confirm.

mscastanho (approved):
@lucmaga LGTM. Only one nit:
I couldn't find the SHA that is mentioned by the last commit message. I believe it was referring to the commit before that one. The SHA may have changed during some rebase you did locally. Just replacing that with `the previous commit` is fine for me, and you could push with that change without further review.

tuliom (approved):
LGTM.
Notice the small fix that Matheus pointed out.

PR #2003, opened 2021-03-08
Update valgrind on AT next

Cherbot:
Bump to revision c79180a3afcf


er-1:
On power, valgrind tests failed with:
```
valgrind: power9:  library /home/tcbot/bot-slave/at_opt/at-next-15.0-0-alpha/lib/valgrind/vgpreload_core-ppc64le-linux.so  is missing
valgrind: power9:  library /home/tcbot/bot-slave/at_opt/at-next-15.0-0-alpha/lib/valgrind/vgpreload_drd-ppc64le-linux.so  is missing
valgrind: power9:  library /home/tcbot/bot-slave/at_opt/at-next-15.0-0-alpha/lib/valgrind/vgpreload_dhat-ppc64le-linux.so  is missing
valgrind: power9:  library /home/tcbot/bot-slave/at_opt/at-next-15.0-0-alpha/lib/valgrind/vgpreload_helgrind-ppc64le-linux.so  is missing
valgrind: power9:  library /home/tcbot/bot-slave/at_opt/at-next-15.0-0-alpha/lib/valgrind/vgpreload_massif-ppc64le-linux.so  is missing
valgrind: power9:  library /home/tcbot/bot-slave/at_opt/at-next-15.0-0-alpha/lib/valgrind/vgpreload_memcheck-ppc64le-linux.so  is missing
```

mscastanho:
Checking the valgrind test, looks like those are *all* the libraries it expects in the output binary.

Going through the list of commits that went in, this is the only that jumps to the eye:
```
commit c9d1cd0db3a806ee0b2e91f59261d6caf791ffee                                                                                                                                                                                                              
Author: Mark Wielaard <mark@klomp.org>                                                                                                                                                                                                                       
Date:   Fri Feb 19 23:49:10 2021 +0100                                                                                                                                                                                                                       

    Use pkglibexec as vglibdir.

    vglibdir is the directory from where valgrind loads its internal tool
    executables and vgpreloads. Currently vglibdir is pkglibdir, so those
    internal tools are intermingeled with normal executables and libraries
    that the user might use directly.

    Make vglibdir equal to pkglibexecdir so the internal tools get installed
    and loaded from libexec and don't get get stored under lib.

    This leaves just the static archives and the mpiwrapper libraries that
    the user would link/load themselves under pkglibdir.

    This seems more in line with the FHS lib/libexec standard and makes it
    slightly easier to combine the tools from a multilib target (say the
    memcheck-amd64-linux and memcheck-x86-linux tools) because they would
    be installed under the same directory, while the pkglibdir can differ
    depending on arch/target (lib/lib64).

    https://bugs.kde.org/show_bug.cgi?id=433323
```

mscastanho:
Looks like these libs are now being installed in `${atdir}/libexec/valgrind`, so the test has to be changed.

mscastanho:
Testing a fix now...

PR #2007, opened 2021-03-10
next: Fix gcc stage 3 build using latest glibc

mscastanho:
This fixes the problems seen with latest glibc update (#1979). Detailed information in the commit messages.

After the dynamic loader fix, gcc_4 started to fail, so I needed to also bump the GCC version to get a fix from upstream. But I'm afraid that will cause us to hit the same problems we saw in #1977 and #1990. At least the native builds should be all passing with this PR.

Closes #1979
Closes #1990


tuliom (approved):
LGTM.

@mscastanho Have you tried to build this on RHEL 8 on POWER9?
I don't think your patch helps issue #1298, but it would be interesting to confirm the problem is still there.

mscastanho:
@tuliom I have not, but I'll run a build to check.

mscastanho:
@tuliom it does not fix #1298. Build failed on configure step for `gcc_3`. From `config.log`:
```
configure:4605: checking whether we are cross compiling
configure:4613: /home/mscastanho/AT/gcc3-with-at-glibc/opt/at-next-15.0-0-alpha/bin/gcc -o conftest -O2 -mminimal-toc -Wl,--dynamic-linker=/home/mscastanho/AT/gcc3-with-at-glibc/opt/at-next-15.0-0-alpha/lib64/ld64.so.2  -L/home/mscastanho/AT/gcc3-with-at-glibc/opt/at-next-15.0-0-alpha/tmp/gcc_3 conftest.c  >&5
configure:4617: $? = 0
configure:4624: ./conftest
./conftest: /lib64/power9/libc.so.6: version `GLIBC_2.34' not found (required by ./conftest)
```
So it is still preferring the CPU-optimized glibc than AT's.

tuliom:
@mscastanho I'm surprised to see the cross compiler errors on gcc_1.
I was not expecting this modification on them too. Have you already taken a look?

mscastanho:
@tuliom gcc_1 is failing on cross targets because we had to update the GCC revision, and then we hit #2017.

mscastanho:
The latest GCC revision fixed #2017, so I rebased against the latest revision and removed the patch to update GCC, which is not needed anymore. Also updated glibc revision to the latest.

tuliom (approved):
LGTM.

PR #2008, opened 2021-03-11
Update valgrind on AT next and fix FVTR test

mscastanho:
Bump to revision c79180a3afcf

Also, update FVTR test to account for the new installation directory for
valgrind's internal libs, which changed to ${atdir}/libexec/valgrind since
upstream commit:

         c9d1cd0db Use pkglibexec as vglibdir.

Closes #2003


mscastanho:
I forgot the tests are shared by all AT versions, I'll have to make sure the new path only applies to AT next and later.

mscastanho:
`AT_PACKAGER_VER` for valgrind on AT next is set to `master`, which breaks the regex that extracts major/minor version information on `fvtr/valgrind/valgrind.exp`, so the test failed on AT next builds

mscastanho:
The only test that failed was a Python FVTR test on AT 12. Most likely #1873. Ok to merge?

rzinsly (approved):

er-1 (approved):

PR #2032, opened 2021-03-24
Disable follow-debug-links in binutils on AT next

er-1:
Since [c46b706620eb](https://sourceware.org/git/?p=binutils-gdb.git;a=commit;h=c46b706620eb), binutils has provided the option follow-debug-links (enabled by default).
This has been included in AT with commit c8efa1662541.

Since that, the tests `ck_ldds` and `systemtap` have failed when ran against the AT packages.

Disabling the option to allow the tests to pass.

### DIFF
--- a/configs/next/packages/binutils/stage_2
+++ b/configs/next/packages/binutils/stage_2
@@ -48,7 +48,8 @@ atcfg_configure() {
 						--enable-shared \
 						--enable-plugins \
 						--enable-install-libiberty=./ \
-						--enable-gold
+						--enable-gold \
+						--enable-follow-debug-links=no
 	else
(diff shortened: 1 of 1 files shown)


mscastanho (approved):
I'm not quite sure if we could hit unexpected side effects from this, now that following the symlinks is the default.

But since with the new flag we basically keep the same behavior as before, and all tests are passing, I vote for merging. We still have time until the release to monitor for bugs coming from this (if there are any).

tuliom (approved):
I think we need further analysis on this to better understand why these errors are happening (with a new issue?).
But that shouldn't block this patch, which keeps the old behavior.

PR #2033, opened 2021-03-25
Fix postinst scriptlet on AT >= 14.0

er-1:
The main loop of the postinst scriptlet of the runtime-debuginfo package
wasn't properly closed.

Added an end to the loop.

### DIFF
--- a/configs/14.0/specs/monolithic.spec
+++ b/configs/14.0/specs/monolithic.spec
@@ -313,6 +313,7 @@ for DIR in $(find %{_prefix}/lib/debug/%{_prefix}/*/ -type d \
     for FILE in $(ls -1 ${DIR}/*.debug 2>/dev/null); do
         ln -sf ${FILE} ${dest}/.debug/$(basename ${FILE} .debug)
     done
+done
 #---------------------------------------------------
 %post devel-debuginfo
 # Create symlinks to the debuginfo files in .debug subdirectories so valgrind,


mscastanho (approved):
LGTM

er-1:
It failed on Debian 10 because the `perf` test didn't pass:
```
erf: power9:  Start logging...
perf: power9:  Attempt to create a probe using a variable name...
perf: power9:  probe-definition(0): test_perf_debug=main argc 
symbol:main file:(null) line:0 offset:0 return:0 lazy:(null)
parsing arg: argc into argc
1 arguments
Open Debuginfo file: /home/tcbot/bot-slave/at_opt/at14.0/bin/ld
Try to find probe point from debuginfo.
Matched function: main [5a5da]
Probe point found: main+0
Searching 'argc' variable in context.
Converting variable argc into trace event.
argc type is int.
Found 1 probe_trace_events.
Opening /sys/kernel/debug/tracing//uprobe_events write=1
Parsing probe_events: p:probe_ld/test_perf_debug /opt/at14.0/bin/ld:0x0000000000006f08 argc=%gpr3:s32
Group:probe_ld Event:test_perf_debug probe:p
Error: event "test_perf_debug" already exists.
 Hint: Remove existing event by 'perf probe -d'
       or force duplicates by 'perf probe -f'
       or set 'force=yes' in BPF source.
  Error: Failed to add events. Reason: File exists (Code: -17)
```
The build machine is shared among people, the existing event is likely to have been created by someone else. We might update the test to handle such case.

That said, I think it's safe to merge the PR.

ThinkOpenly:
> It failed on Debian 10 because the `perf` test didn't pass:
> 
> ```
> Parsing probe_events: p:probe_ld/test_perf_debug /opt/at14.0/bin/ld:0x0000000000006f08 argc=%gpr3:s32
> Group:probe_ld Event:test_perf_debug probe:p
> Error: event "test_perf_debug" already exists.
> ```
> 
> The build machine is shared among people, the existing event is likely to have been created by someone else. We might update the test to handle such case.

I have changes in hand for such an update to the test, and I've created issue #2047 to track.

PR #2135, opened 2021-05-12
Update speccpu2017 config files

tuliom:
Newer GCC versions started to cause build or runtime errors on SPEC
CPU2017, requiring extra flags.

Besides that, remove unnecessary linker parameters.


er-1:
All the buils passed.
AT next Ubuntu20 ppc64le 642 was rebuilt and passed
The pending builds are a github glitch, they were rebuilt and passed.

PR #2149, opened 2021-05-22
Update glibc on AT next

Cherbot:
Bump to revision 49e182c7376b

PR #2154, opened 2021-05-24
fvtr: Use proper paths to check for SystemTap probes

mscastanho:
Starting with glibc 2.34 (AT 15.0), libpthread has been moved into libc.so, so
`fvtr/systemtap/systemtap.exp` cannot keep checking for SDT probes on libpthread,
but instead should use libc.so. This commit fixes that test to work with both
older and newer AT versions.

This also includes the glibc version update from #2149 to make sure the issue is fixed.

Closes #2149


fvtr/systemtap/systemtap.exp
```diff
@@ -49,17 +49,21 @@ if { $env(AT_CROSS_BUILD) == "yes" } {
...
-		set ret 1
-	}
+
+if { $::env(AT_MAJOR_VERSION) < 15.0 } {
+	set lib [glob $libpath/libpthread-*.so]
+} else {
+	set lib [glob $libpath/libc-*.so]
+}
+
+if {[check_systemtap $readelf_exe $lib]} {
```
tuliom, line 65: I don't think these changes are correct because they assume a single libc or libpthread is available.
In a big endian system, AT might provide both 32 and 64 bits libraries.  That means the test has to be executed twice.

fvtr/systemtap/systemtap.exp
```diff
@@ -49,17 +49,21 @@ if { $env(AT_CROSS_BUILD) == "yes" } {
...
+} else {
+	set libpath "$install_prefix/lib64"
 }
-if { $TARGET64 } {
-	set lib [glob $install_prefix/lib64/libpthread-*.so]
-	if {[check_systemtap $readelf_exe $lib]} {
-		set ret 1
-	}
+
+if { $::env(AT_MAJOR_VERSION) < 15.0 } {
```
tuliom, line 59: Can we test for `AT_GLIBC_VER < 2.34` instead?
That works even if someone decides to build an AT 14.0 with glibc 2.34.

tuliom (changes requested):

fvtr/systemtap/systemtap.exp
mscastanho: Sure. Changed in the latest push.

fvtr/systemtap/systemtap.exp
mscastanho: True. I changed the PR to maintain the same test structure as before (guaranteeing the test will run twice, for 32-bit and 64-bit).

tuliom (approved):
LGTM.
Please wait for tests to complete/pass before merging this.

mscastanho:
The test worked for older AT versions, but failed on next because glibc 2.34 has not been released yet, so the version information is set to `2.33.9000`, so we can't compare against 2.34, at least not until it is released, when 2.33.9000 will be extinguished.

mscastanho:
The latest changes should handle that.

mscastanho:
@tuliom are you also OK with this new version?

PR #2145, opened 2021-05-18
Enable auto-merge for automated package update PRs

mscastanho:
Improve the script that creates our automated pull requests to also add an
'auto' tag to the PR. This will help our CI know which PRs are candidates for
auto-merging in the future.

PR #2155, opened 2021-05-25
Enable bumping version

er-1:
Allow the script that creates our automated pull requests to bump the version in addition to bump the revision.
If a new version version is found, it'll disable the auto-merge.


scripts/utilities/update_revision.sh
```diff
@@ -514,9 +591,11 @@ for co in "${ATSRC_PACKAGE_CO[@]}"; do
 	if [[ -n "${hash}" ]]; then
 	    echo ""
 	    print_msg 0 "The latest revision of ${repo} is ${hash}"
-	    update_revision ${1} ${hash}
+
+	    version=$(get_version ${package} ${hash})
```
mscastanho, line 595: What happens if for some reason we reach the `*)` case on `get_version`? If I understood correctly, `$version` here wouldn't be null, but contain something like `"Function get_version doesn't know the package $package"`, which will  trigger some code paths on `send_to_github`.

It would be good to check the return value of `get_version` and only set `$version` here if no error occurred.

mscastanho (commented):
@er-1 Overall looks good. 
Adding the `auto` label as I did on PR #2145 is probably not going to be necessary anymore. I'm still working on that. I think you could rebase your patch on top of master. 

Also, see my comment below.

scripts/utilities/update_revision.sh
er-1: Done.

er-1:
> Adding the `auto` label as I did on PR #2145 is probably not going to be necessary anymore. I'm still working on that. I think you could rebase your patch on top of master.
@mscastanho The idea is to prevent the auto-merge when a new version is found. Currently, it skips adding the `auto` label in that case. I'll update my patch after your changes.

tuliom (approved):

PR #2164, opened 2021-06-01
Update expat on AT next

er-1:
Commit [779d147681f4](https://github.com/libexpat/libexpat/commit/779d147681f4ff3bf399f5c4f2d77a8f0d266db4) updated the buildconf.sh script of expat. Because of the changes, the script can't be run by dash. However, on Debian and Ubuntu, sh is a symbolic link to dash. So calling the buildconf.sh script using sh leads to an error on Debian and Ubuntu.

Also:
Bump to version 2.4.1
Bump to revision a28238bdeebc
    
Close #2158 

Signed-off-by: Erwan Prioul <erwan@linux.ibm.com>

PR #2165, opened 2021-06-01
Clone AT next into AT 15.0

er-1:
Prepare AT 15.0 based on AT next for further changes.


mscastanho (approved):
LGTM. Of course I didn't do a detailed inspection of all 342 files, but I tried to check the following things:
- AT 15 files are symlinks  to the AT 14 ones where possible
- The files that are not symlinks are either `sources` or needed different flags/changes compared to AT14
- next files now point to AT15's

Looks like the PR respects all of the above.

mscastanho:
I suspect it would be better to merge #2164 first and then to a small rebase to get the changes from that.

Issue #2204, opened 2021-06-24
go compiler on AT14.0 on POWER10 is broken


carlelove:
I am trying to compile a go program from the  GDB testsuite and the go compiler segfaults.

Here is how to reproduce the issue.

1) on ltcden2-lp1 cd to the GDB go tests.  In my case they are in carll@ltcden2-lp1:~/GDB/binutils-gdb-current/gdb/testsuite/gdb.go
2) try to compile the intefers.go file, i.e.

     carll@ltcden2-lp1:~/GDB/binutils-gdb-current/gdb/testsuite/gdb.go$ go integers.go
     Segmentation fault (core dumped)

tuliom:
It looks like there is something wrong with `/opt/at14.0/lib64/power10/libgo.so.16`:

```
$ /opt/at14.0/bin/go help
Segmentation fault (core dumped)
$ LD_PRELOAD=/opt/at14.0/lib64/libgo.so.16 /opt/at14.0/bin/go help
Go is a tool for managing Go source code.
...
```

It's stuck in a function that I can't identify:

```
=> 0x00007ffff7726f80 <+64>:    ld      r11,0(r3)
   0x00007ffff7726f84 <+68>:    ld      r12,8(r3)
   0x00007ffff7726f88 <+72>:    mr      r0,r3
   0x00007ffff7726f8c <+76>:    cmpdi   r11,0
   0x00007ffff7726f90 <+80>:    add     r3,r12,r13
   0x00007ffff7726f94 <+84>:    beqlr
   0x00007ffff7726f98 <+88>:    mr      r3,r0
   0x00007ffff7726f9c <+92>:    mflr    r11
   0x00007ffff7726fa0 <+96>:    std     r11,8(r1)
   0x00007ffff7726fa4 <+100>:   std     r2,24(r1)
   0x00007ffff7726fa8 <+104>:   ld      r12,4744(r2)
   0x00007ffff7726fac <+108>:   mtctr   r12
   0x00007ffff7726fb0 <+112>:   bctrl
   0x00007ffff7726fb4 <+116>:   ld      r2,24(r1)
   0x00007ffff7726fb8 <+120>:   ld      r11,8(r1)
   0x00007ffff7726fbc <+124>:   mtlr    r11
   0x00007ffff7726fc0 <+128>:   blr
```

This function received an `r3` value that looks broken:
```
>>> print/x $r3
$1 = 0xf20
```

[...]

PR #2214, opened 2021-06-30
Add a patch to fix issue #2204

tuliom:
There has been an issue in GCC 10 whose fix depend on a patch already
available in GCC 11 and a patch for GCC 12 that are under review.

I've already asked for both patches to be backported.

Meanwhile, we may need to carry an extra patch in GCC 10.

Signed-off-by: Tulio Magno Quites Machado Filho <tuliom@linux.ibm.com>


tuliom:
The build failed when creating the deb package:
```
perl debian/dh_strip --dest=/home/tcbot/bot-slave/at_opt/at14.0 --exclude=getconf --exclude=libgo. --exclude=ld- --dbg-package=advance-toolchain-at14.0-runtime-dbg -p advance-toolchain-at14.0-runtime
make[1]: *** [debian/rules:40: binary-arch] Error 1
```

tuliom:
I haven't been able to reproduce the issue with semop.
Anyway, the issue looks unrelated to this change.

mscastanho (approved):

PR #2215, opened 2021-07-01
Improve checking version of packages

er-1:
Moved into the sources files the way to get the version of the package
upstream. This allows to follow changes between the branches of a package (like liburcu for example).

### DIFF
--- a/configs/13.0/packages/binutils/sources
+++ b/configs/13.0/packages/binutils/sources
@@ -41,6 +41,7 @@ ATSRC_PACKAGE_PATCHES=
 ATSRC_PACKAGE_TARS=
 ATSRC_PACKAGE_DISTRIB=yes
 ATSRC_PACKAGE_BUNDLE=main_toolchain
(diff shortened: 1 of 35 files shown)


mscastanho (changes requested):
@er-1 A few questions and comments:
- I don't see any updates to `libhugetlbfs` files, even though it's version was being checked by `update_revision.sh` before.
- Missing `ATSRC_PACKAGE_UPSTREAM` for liburcu and valgrind on AT next

For the last point, I believe it's because both are pointing to the master branch, so would be ignored by the original script, right? But could you add that line commented out to those files? Just to document why they are not needed there. Something like:

```
# Not needed because file follows master branch
# ATSRC_PACKAGE_UPSTREAM= 
```

er-1:
> @er-1 A few questions and comments:
> 
>     * I don't see any updates to `libhugetlbfs` files, even though it's version was being checked by `update_revision.sh` before.

It's because it's pointing to the master branch. The check in `update_revision.sh` was ignored.

>     * Missing `ATSRC_PACKAGE_UPSTREAM` for liburcu and valgrind on AT next
> 
> For the last point, I believe it's because both are pointing to the master branch, so would be ignored by the original script, right? 

Right.

> But could you add that line commented out to those files? Just to document why they are not needed there. Something like:
> 
> ```
> # Not needed because file follows master branch
> # ATSRC_PACKAGE_UPSTREAM= 
> ```

Good idea.

mscastanho (approved):

tuliom (approved):

PR #2225, opened 2021-07-05
Update glibc on AT next and fix FVTR test

mscastanho:
Bump to revision e766ce3088e1

Starting with glibc commit '8208be389b * Install shared objects under their ABI
names', glibc does not install libc-<version>.so files anymore, instead
libc.so.6 became a regular file (was a symlink before).

This breaks systemtap.exp because it looks for the old-style .so name. This
commit updates the test to make sure it uses the correct .so name for glibc >=
2.34.

Close #2219


tuliom:
"AT 14.0 Ubuntu20 ppc64le" failed because the disk was full.
The disk has been cleaned up and I requested another build.

er-1:
Failed on Ubuntu 20.04 while creating the `deb` files:
```
dh_systemd_enable
dh_install
dh_missing --fail-missing
make[1]: *** [debian/rules:38: binary-arch] Error 1
make[1]: Leaving directory '/home/tcbot/bot-worker/at/src/at16.0-0-alpha.ubuntu-20_ppc64le_ppc64le/debs/debhelper'
/bin/fakeroot: 1: kill: No such process
```

PR #2227, opened 2021-07-06
Allow auto update for some packages

er-1:
Use git repository instead of tarball to get source code, it allows the update_revision.sh script to get upstream updates.


configs/next/packages/libnxz/sources
```diff
@@ -20,17 +20,17 @@
 #
 
 ATSRC_PACKAGE_NAME="NX GZIP library"
-ATSRC_PACKAGE_VER="v0.61"
+ATSRC_PACKAGE_VER=master
```
rzinsly: On libnxz "master" is the stable branch, we should track "develop" for AT next.

rzinsly (commented):
For some reason github is not letting me comment in the openssl commit:

Is it intentional to update OpenSSL on AT 12.0 from 1.1.0 to 1.1.1?
Also, is it safe to stop following the stable branches?

configs/next/packages/libnxz/sources
er-1: Fixed.

er-1:
> Is it intentional to update OpenSSL on AT 12.0 from 1.1.0 to 1.1.1?

No. My bad.
Fixed.

configs/next/packages/libnxz/stage_1
```diff
@@ -37,11 +37,6 @@ atcfg_pre_configure() {
 atcfg_configure() {
 	# libnxz do not use autoconf
 	${SUB_MAKE} clean
-	# Hack to get the version outside the git repository.
-	cat lib/Makefile | \
-		sed "s/git describe --tags | cut -d - -f 1,2 | tr - ./echo ${ATSRC_PACKAGE_VER}/" \
-		> lib/Makefile.temp
-        [[ ${?} -eq 0 ]] && mv lib/Makefile.temp lib/Makefile || exit 1
```
rzinsly: We will still need this hack because the git repository will be on `sources` and the build is done on `$AT_WD/sources`, that won't have the `.git` directory and files.

rzinsly (changes requested):
It is not necessary for this patch, but for future patches you could have all "Replace links" commits squashed in a single commit.

mscastanho (commented):
Just to be sure: so this hack for libnxz is not needed anymore?
https://github.com/advancetoolchain/advance-toolchain/blob/4057e4f15a02d94f6984110c855940dc44ea4319/configs/14.0/packages/libnxz/stage_1#L40-L44

mscastanho:
Oops, looks like @rzinsly was faster than me =)

tuliom:
I'm afraid commit 48831c3c123170fbb14c20fe5f15996d4bc1d8d7 is changing more files than expected, I expected only the symlink replacement for paflib, but it's also undoing part of the changes for libdfp.

tuliom:
zlib uses a different branch naming. Development happens on branch `develop`.
By using the master branch, you'd only get stable releases, which looks good to me.

tuliom:
Why are you removing the revision from `ATSRC_PACKAGE_STR_VER`?

configs/next/packages/libnxz/stage_1
er-1: @rzinsly @mscastanho My understanding is this hack is needed when we take the sources from the tarball as AT <= 15.0 does. But AT next will `git clone` the sources, so `git` commands will be available and then the hack won't be necessary.

er-1:
> I'm afraid commit [48831c3](https://github.com/advancetoolchain/advance-toolchain/commit/48831c3c123170fbb14c20fe5f15996d4bc1d8d7) is changing more files than expected, I expected only the symlink replacement for paflib, but it's also undoing part of the changes for libdfp.

My bad. I'll fix that.

er-1:
> Why are you removing the revision from `ATSRC_PACKAGE_STR_VER`?

You're right, no need for this change.
Fixed.

configs/next/packages/libnxz/stage_1
rzinsly: @er-1 `lib/Makefile` on libnxz uses `git describe --tags` on line 6, that command won't be available at build time because the source copied to `$AT_WD/sources` is not a git repository, you only are able to run that in the `sources` at the top directory, there is where the `git clone` is done.

er-1: > `lib/Makefile` on libnxz uses `git describe --tags` on line 6, that command won't be available at build time because the source copied to `$AT_WD/sources` is not a git repository

@rzinsly The `.git` directory isn't copied with the whole sources? I thought the copy was done using `cp -r`, so `.git`would be copied making the directory under `$AT_WD/sources` a git repository.

rzinsly: @er-1 the copy is done with rsync and we explicit exclude all version control files with `--exclude=CVS --exclude=.svn --exclude=.git`

er-1: Fixed.

rzinsly (approved):

PR #2231, opened 2021-07-08
Fix stderr redirections

tuliom:
Stop discarding output sent to stderr and properly forward them to the
same log file used for stdout.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -351,7 +351,7 @@ define runandlog
       echo; \
       date; \
       echo "-----------------------------------------------------"; \
-    } >> $1 2>&1
+    } &>> $1
 endef
 
 # Create path if needed and return its name
@@ -1017,7 +1017,7 @@ pack: hash
 	@echo "$$($(TIME)) Preparing the build pack..."
 	@+{ TMP_DIR=$$(mktemp -d "/tmp/atpack-XXXXXXX"); \
 	    tar -cpvz -X "$(HELPERS_ROOT)/pack-exclude.lst" -f "$${TMP_DIR}/at-$(AT_TODAY).tgz" ./ > \
-	        "$${TMP_DIR}/at-$(AT_TODAY).included" 2>&1; \
+	        "$${TMP_DIR}/at-$(AT_TODAY).included"; \
 	    mv "$${TMP_DIR}/at-$(AT_TODAY).tgz"      . ; \
 	    mv "$${TMP_DIR}/at-$(AT_TODAY).included" . ; \
 	    rm -rf "$${TMP_DIR}"; \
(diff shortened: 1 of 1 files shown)


ThinkOpenly (changes requested):
need matching changes on lines 563, 991, 1009, 1024, 1140, 1170, 1190, 1218, 1363, 1370, 1377, 1420?

tuliom:
> need matching changes on lines 563, 991, 1009, 1024  ?

@ThinkOpenly I intentionally left them out because I think these errors are being collected somewhere.
If I change them, they would be directed to `/dev/null`.
Do you think this a good idea? It was the intention of the original author.

> 1140, 1170, 1190, 1218, 1363, 1370, 1377, 1420.

Oops.  I'm fixing them.

Thanks!

ThinkOpenly:
> > need matching changes on lines 563, 991, 1009, 1024  ?
> 
> @ThinkOpenly I intentionally left them out because I think these errors are being collected somewhere.
> If I change them, they would be directed to `/dev/null`.
> Do you think this a good idea? It was the intention of the original author.

Ah! I see. Let's leave them as is.

tuliom:
I fixed the issues.  I also found an extra place that was missing in my previous patch.

Makefile
```diff
@@ -1016,8 +1016,8 @@ edit:
 pack: hash
 	@echo "$$($(TIME)) Preparing the build pack..."
 	@+{ TMP_DIR=$$(mktemp -d "/tmp/atpack-XXXXXXX"); \
-	    tar -cpvz -X "$(HELPERS_ROOT)/pack-exclude.lst" -f "$${TMP_DIR}/at-$(AT_TODAY).tgz" ./ > \
-	        "$${TMP_DIR}/at-$(AT_TODAY).included" 2>&1; \
+	    tar -cpvz -X "$(HELPERS_ROOT)/pack-exclude.lst" -f "$${TMP_DIR}/at-$(AT_TODAY).tgz" ./ &> \
+	        "$${TMP_DIR}/at-$(AT_TODAY).included"; \
```
ThinkOpenly, line 1020: I'm not sure about this one. I think the intent here is (or should be) to let stderr float until being picked up at line 1024. Maybe instead of `&>`, it should just be `>` and remove the `2>&1`?

ThinkOpenly (changes requested):

ThinkOpenly (changes requested):
Sorry, but what do you think about lines 1296, 1322?  I think those need `s/>/&>/`.

tuliom:
> Sorry, but what do you think about lines 1296, 1322? I think those need `s/>/&>/`.

Yes! I've just pushed these changes and I also modified line 1337.

Thanks!

tuliom:
`AT 11.0 Debian9 ppc64le` failed due to an error when testing python.  The failure isn't related to this PR.

ThinkOpenly (approved):

PR #2216, opened 2021-07-02
Update GCC on AT next and apply fixes

mscastanho:
Bump to revision 4546f423ecff

We haven't been able to update GCC on AT next since GCC commit 'ebd5e86c0f4
Improve global state for options', which is causing issues when building
glibc. The issue has already been reported upstream (Bug 100909) [1], but hasn't
been fixed yet. So meanwhile, let's revert it off-tree.

Also, since '6cf0040fff7 libstdc++: Improve std::lock algorithm' gcc_1 fails to
build because libstdc++ tries to call '__gthread_yield' even when we compile
with --disable-threads. A patch has been submitted upstream [2] and has been
approved, but is still waiting to be merged. Meanwhile, apply it off-tree as
well.

[1] https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100909
[2] https://gcc.gnu.org/pipermail/libstdc++/2021-June/052858.html

Close #2159

PR #2234, opened 2021-07-09
Move AT 15.0 to python branch 3.9

er-1:
AT 15.0 will track the 3.9 branch, whereas AT next will track the 3.10 branch.


er-1:
AT next failed while building pveclib.
AT next is using the same pveclib as AT 15.0. Since 2b86315ea5051e5486730e342e, AT next has failed to build. PR #2216 should fix it.

mscastanho:
I was surprised that even though the `3.10` branch does not contain commit `91db097358bc` (from `next/packages/python/sources`) the build did not fail.

er-1:
> I was surprised that even though the `3.10` branch does not contain commit `91db097358bc` (from `next/packages/python/sources`) the build did not fail.

That's a trick.
When building, it takes only the revision in account.
`update_revision.sh` will bump the revision (and the version) based on the branch.

mscastanho:
Oh, nice. Good to know. Thanks for the explanation.

mscastanho (approved):
LGTM.

PR #2242, opened 2021-07-12
Fix python build for 32 bit version

lucmaga:
Python build fails to find the correct openssl when building on 32
bit systems.

Signed-off-by: Lucas A. M. Magalhaes <lamm@linux.ibm.com>

### DIFF
--- a/configs/13.0/packages/python/stage_1
+++ b/configs/13.0/packages/python/stage_1
@@ -46,7 +46,7 @@ atcfg_configure() {
 		--libdir="${at_dest}/lib${compiler##32}" \
 		--enable-shared \
 		--enable-loadable-sqlite-extensions \
-		--with-openssl="${at_dest}" \
+		--with-openssl="${at_dest}/lib${compiler##32}" \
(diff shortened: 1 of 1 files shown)


tuliom:
It looks like the python tests failed because the SSL module wasn't built.

Looking at the documentation for this parameter, it says:
```
--with-openssl=DIR      root of the OpenSSL directory
```

Which I interpret as a directory that doesn't contain `/lib`, e.g. `/opt/at13.0` instead of `/opt/at13.0/lib64`.
Looking at the default values, they also point to directories without `/lib`:
```
ssldirs="/usr/local/ssl /usr/lib/ssl /usr/ssl /usr/pkg /usr/local /usr"
```

Anyway, I don't understand the issue you're trying to fix.
If you're using a pure 32-bit system, I wouldn't be surprised if many more issues would appear before the build reaches python.
If you're using a 64-bit big endian system, I expected Python would be 64 bits and it wouldn't use the 32-bit OpenSSL libraries.

PR #2249, opened 2021-07-17
Update python on AT next

Cherbot:
Bump to version 4
Bump to revision efda9054b9fc


mscastanho:
This PR bumped the Python version to 4. But looking at the official website, git logs, news, I couldn't find any mention to Python 4. Maybe the script messed up this time?

mscastanho:
Also failed when building boost.python:
```
gcc.compile.c++ /home/tcbot/bot-worker/at/src/at16.0-0-alpha.suse-15_ppc64le_ppc64le/builds/boost_1-64/build/boost/bin.v2/libs/python/build/gcc-12.0.0/release/link-static/python-2.7/threading-multi/visibility-hidden/list.o
In file included from ./boost/python/detail/prefix.hpp:13,
                 from ./boost/python/list.hpp:8,
                 from libs/python/src/list.cpp:5:
./boost/python/detail/wrap_python.hpp:82:10: fatal error: patchlevel.h: No such file or directory
   82 | #include <patchlevel.h>
      |          ^~~~~~~~~~~~~~
compilation terminated.
```

mscastanho:
This is the command `update_revision.sh` is currently using to get the latest Python version:
```
$ wget -qO - "https://raw.githubusercontent.com/python/cpython/efda9054b9fc/README.rst" | head -n1 | grep "This is Python version " | sed "s/^.* \([0-9\.][0-9\.]*\).*$/\1/"                                           
4
```
This is happening because the version in the README changed to `3.10.0 beta 4`:
```
$ wget -qO - "https://raw.githubusercontent.com/python/cpython/efda9054b9fc/README.rst" | head -n1 | grep "This is Python version "                                         
This is Python version 3.10.0 beta 4
```

and the regex is getting only the last string, which is `4` in this case. This slightly changed regex seems to fix the issue:
```
$ wget -qO - "https://raw.githubusercontent.com/python/cpython/efda9054b9fc/README.rst" | head -n1 | grep "This is Python version " | sed "s/^.* version \([0-9\.][0-9\.]*\).*$/\1/"
3.10.0
```
@er-1 Do you think this is OK or do we actually need the info about beta/alpha version on `ATSRC_PACKAGE_VER` as well?

er-1:
> @er-1 Do you think this is OK or do we actually need the info about beta/alpha version on `ATSRC_PACKAGE_VER` as well?

@mscastanho 
This is OK without alpha/beta info.
Thanks for taking care of this.

mscastanho:
@er-1 no problem!
`ck_ldds` fvtr test also failed but I think it was a side-effect of the bump to version 4.

er-1:
> `ck_ldds` fvtr test also failed but I think it was a side-effect of the bump to version 4.

I agree.

I think the build #112 could be cancelled, this PR is not going to be merged, so there is no need to wait for the build on RHEL8 ppc64le.

mscastanho:
> I think the build 112 could be cancelled, this PR is not going to be merged, so there is no need to wait for the build on RHEL8 ppc64le.

Agreed. I just canceled it.

mscastanho:
Closing to let Cherbot open a new PR with the updated regex from #2254 during the weekend.

Issue #1969, opened 2021-02-16
openssl: Missing EVP_KDF symbols on AT 14.0


tuliom:
An user reported the following missing symbols from libcrypto on AT 14.0:

```
/opt/at14.0/lib/gcc/powerpc64le-linux-gnu/10.2.1/../../../../powerpc64le-linux-gnu/bin/ld: /lib64/libk5crypto.so.3: undefined reference to `EVP_KDF_derive@OPENSSL_1_1_1b'
/opt/at14.0/lib/gcc/powerpc64le-linux-gnu/10.2.1/../../../../powerpc64le-linux-gnu/bin/ld: /lib64/libk5crypto.so.3: undefined reference to `EVP_KDF_ctrl@OPENSSL_1_1_1b'
/opt/at14.0/lib/gcc/powerpc64le-linux-gnu/10.2.1/../../../../powerpc64le-linux-gnu/bin/ld: /lib64/libk5crypto.so.3: undefined reference to `EVP_KDF_CTX_new_id@OPENSSL_1_1_1b'
/opt/at14.0/lib/gcc/powerpc64le-linux-gnu/10.2.1/../../../../powerpc64le-linux-gnu/bin/ld: /lib64/libk5crypto.so.3: undefined reference to `EVP_KDF_CTX_free@OPENSSL_1_1_1b'
```

tuliom:
These functions are available on OpenSSL master branch [for version 3.0](https://github.com/openssl/openssl/blob/master/NEWS.md#major-changes-between-openssl-111-and-openssl-30-under-development).

But Fedora decided to [backport them to OpenSSL 1.1.1](https://src.fedoraproject.org/rpms/openssl/c/5cda1ca091fe7a96730a929b2a0333aabb6b3966) breaking the ABI with upstream builds because they're providing symbols that are not available upstream.

Looking at `abidiff` I noticed the following `EVP_KDF` symbols added to RHEL 8:
```
  EVP_KDF_CTX_free@@OPENSSL_1_1_1b
  EVP_KDF_CTX_new_id@@OPENSSL_1_1_1b
  EVP_KDF_ctrl@@OPENSSL_1_1_1b
  EVP_KDF_ctrl_str@@OPENSSL_1_1_1b
  EVP_KDF_derive@@OPENSSL_1_1_1b
  EVP_KDF_reset@@OPENSSL_1_1_1b
  EVP_KDF_size@@OPENSSL_1_1_1b
[...]

PR #2253, opened 2021-07-18
Enable AT builds on RHEL8

### LINKED ISSUE #1298: AT fails to build on RHEL 8

ThinkOpenly:
Building on systems with optimized libraries can conflict with the
AT build.  The loader, including the loader that is built with AT,
will prefer libraries in so-called "optimized" paths.  These are
any paths which are subdirectories of a path in the `ld.so.conf`
configuration which are classified as "optimized" by a predefined
list, generally by processor architecture, like `power9`.
So, a file in `/lib/power9/libc.so` will take preference over
`/opt/atX/lib/libc.so`, even if `/opt/atX/lib` is earlier than
`/lib` in the `ld.so.conf` search order.

The problem coveniently occurs between the two `ldconfig`
stages.  Work around the problem by creating a fake "optimized"
set of libraries which are simply copied from the "unoptimized"
directory into "optimized" subdirectories of a new (temporary)
directory, `syslib-override`.

This new directory is removed from `ld.so.conf` and the directory
tree is completely removed before the second `ldconfig` runs.

Fixes #1298.

### DIFF
--- a/scripts/helpers/utilities.mk
+++ b/scripts/helpers/utilities.mk
@@ -364,6 +364,19 @@ endef
 # still distributed in /lib or /usr/lib instead of the correct multiarch
 # directory.  So, it's necessary to carry a hack until they complete to port
 # all packages to multiarch.
+#
+# Another hack below is the "syslib-override" directory.
+# During the build, there is an awkward period where some packages are
+# built depending on AT libraries, but optimized libraries are not yet
+# built. The loader will always prefer optimized libraries *even if they
+# are under directories lower in the search order*.  For example,
+# /lib64/power9/libc.so is preferred over /opt/atX.0/lib64/libc.so.
+# This causes all sorts of problems during this awkward period.
+# So, the syslib-override directory is put at the top of the search order
+# and populated with AT-built (not-optimized) libraries in optimized
+# subdirectories, so they will be used in preference, and the build
+# can proceed through the awkward period. After that period, the entire
+# syslib-override directory tree is removed.
 define prepare_loader_cache
     set -e; \
     if [[ ! -d "$(AT_DEST)/etc/ld.so.conf.d" ]]; then \
@@ -377,7 +390,11 @@ define prepare_loader_cache
         echo "/usr/local/lib"                 >> "$(DYNAMIC_LOAD)/sys32.ld.conf"; \
     fi; \
     if [[ -n "$(TARGET64)" ]]; then \
-        echo "$(AT_DEST)/$(TARGET)/lib64" >  "$(DYNAMIC_LOAD)/at64.ld.conf"; \
+        > "$(DYNAMIC_LOAD)/at64.ld.conf"; \
+        if [[ $1 -eq 1 ]]; then \
+            echo "$(AT_DEST)/syslib-override"   >  "$(DYNAMIC_LOAD)/at64.ld.conf"; \
+        fi; \
+        echo "$(AT_DEST)/$(TARGET)/lib64" >> "$(DYNAMIC_LOAD)/at64.ld.conf"; \
         echo "$(AT_DEST)/lib64"           >> "$(DYNAMIC_LOAD)/at64.ld.conf"; \
         echo "/lib64"                     >  "$(DYNAMIC_LOAD)/sys64.ld.conf"; \
         if [[ ("$(DISTRO_FM)" == "ubuntu") || ("$(DISTRO_FM)" == "debian") ]]; then \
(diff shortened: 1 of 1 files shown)


scripts/helpers/utilities.mk
```diff
@@ -377,7 +390,10 @@ define prepare_loader_cache
         echo "/usr/local/lib"                 >> "$(DYNAMIC_LOAD)/sys32.ld.conf"; \
     fi; \
     if [[ -n "$(TARGET64)" ]]; then \
-        echo "$(AT_DEST)/$(TARGET)/lib64" >  "$(DYNAMIC_LOAD)/at64.ld.conf"; \
+        if [[ $1 -eq 1 ]]; then \
+            echo "$(AT_DEST)/slib-override"   >  "$(DYNAMIC_LOAD)/at64.ld.conf"; \
+        fi; \
+        echo "$(AT_DEST)/$(TARGET)/lib64" >> "$(DYNAMIC_LOAD)/at64.ld.conf"; \
```
tuliom, line 396: Is there another mechanism removing `$(DYNAMIC_LOAD)/at64.ld.conf` between executions 1 and 2?

I'm asking because I'm afraid this code won't generate correct results if that doesn't happen, i.e. the line with slib-override will still be there and `$(AT_DEST)/$(TARGET)/lib64` will be duplicated.

tuliom (commented):
The code looks good to me in overall.  The source code comment is great!
But I have a question that may lead to more changes.

scripts/helpers/utilities.mk
ThinkOpenly: I took me a minute, but I understand you have noticed that the `>` operation is within the `if` clause, so the file will only be appended during execution 2. Let me make sure that the file is cleared / recreated during the second run.

ThinkOpenly:
Note that the builds on RHEL8 still fail even after the recent force-push. The remaining issues, however, are covered by #1969.

I propose we merge this PR, and then focus on #1969, which has a little more complexity, it would seem.

ThinkOpenly:
Build "AT 11.0 RHEL7 ppc64le" failed:

> 0:01:53 load avg: 2.88 [376/376/1] test_subprocess failed
> test test_subprocess failed -- Traceback (most recent call last):
>   File "/home/tcbot/bot-slave/at_opt/at11.0/lib64/python3.6/test/test_subprocess.py", line 1078, in test_wait_timeout
>     p.wait(timeout=0.0001)
> AssertionError: TimeoutExpired not raised

Looks like another semi-random Python failure, so I'm going to ignore it. The rest of the builds are even more successful than I was expecting. I'll wait for the few remaining to complete...

PR #2254, opened 2021-07-20
next: Fix regex to get current Python version

mscastanho:
The current regex passed to sed on ATSRC_PACKAGE_UPSTREAM for Python on AT next
is not handling beta/alpha versions correctly (see #2249). Change it so that it
ignores any beta/alpha information and only output the main version.


mscastanho:
The same regex is also used to check the version on AT 13, 14 and 15. But since those follow released versions, the old regex behaves correctly, so no need to change it there and trigger other 32 test builds.

er-1 (approved):

tuliom (approved):

PR #2255, opened 2021-07-23
Fix update_revision.sh to handle tarball file

### LINKED ISSUE #2226: `update_revision.sh` treats URL to tar.gz file as SVN URL
On AT 13, we download Python source code as tarball, instead of using the git repo:

https://github.com/advancetoolchain/advance-toolchain/blob/25d72e8236654e4f4c1cb32985f5271ce49b26fd/configs/13.0/packages/python/sources#L32

When `update_revision.sh` parses `ATSRC_PACKAGE_CO`, it detects it as an SVN URL and tries to run `svn` on it:

https://github.com/advancetoolchain/advance-toolchain/blob/25d72e8236654e4f4c1cb32985f5271ce49b26fd/scripts/utilities/update_revision.sh#L205-L208

Which is causing issues when the automated update script is run:
```
Sat Jul  3 01:05:51 EDT 2021 - python 13.0
svn: warning: W170000: URL 'https://github.com/python/cpython/archive/13c94747c744.tar.gz' non-existent in revision 285718

svn: E200009: Could not display info for all targets because some targets don't exist
Unable to connect to https://github.com/python/cpython/archive/13c94747c744.tar.gz or wrong branch name
```

Note this does not happen with later AT versions because their `sources` file set `ATSRC_PACKAGE_CO` to clone the git repo directly.

er-1:
The script handles tarball file as SVN URL causing inappropriate warnings.
This patch adds a check to verify if the given URL is a tarball or not, if not
it assumes it's a SVN URL.

Fix #2226 

### DIFF
--- a/scripts/utilities/update_revision.sh
+++ b/scripts/utilities/update_revision.sh
@@ -203,8 +203,13 @@ get_latest_revision ()
 		popd > /dev/null
 		rm -rf ${tmp_dir}
 	else
-		hash=$(svn info ${url} | grep "Last Changed Rev:" \
-			| cut -d: -f2)
+		local isTgz=$(echo ${url} | grep -c "\.tar\.[bg]z")
+		if [[ "${isTgz}" == 1 ]]; then
+			echo "URL is a tarball, nothing to do."
+			return 1
+		fi
+		# It's not a tarball, let's assume it's a SVN URL
+		hash=$(svn info ${url} | grep "Last Changed Rev:" | cut -d: -f2)
 	fi
 
 	echo ${hash}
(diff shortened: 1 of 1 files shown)


scripts/utilities/update_revision.sh
```diff
@@ -203,8 +203,13 @@ get_latest_revision ()
...
 		rm -rf ${tmp_dir}
 	else
-		hash=$(svn info ${url} | grep "Last Changed Rev:" \
-			| cut -d: -f2)
+		local isTgz=$(echo ${url} | grep -c "\.tar\.[bg]z")
+		if [[ "${isTgz}" == 1 ]]; then
+			echo "URL is a tarball, nothing to do."
+			return 1
+		fi
+		# It's not a tarball, let assume it's a SVN URL
```
mscastanho, line 211: let -> let's

mscastanho (approved):
LGTM. Just a minor typo.

scripts/utilities/update_revision.sh
mscastanho: No need to run the tests again if you push this typo fix.

rzinsly (approved):

PR #2241, opened 2021-07-12
Update GCC on AT next and remove out-of-tree patch

mscastanho:
Bump to revision 2e345e4ad6bb

The libstdc++ gthreads fix has been merged upstream (commit 2e345e4ad6bb), so we
don't need to apply it anymore.

PR #2258, opened 2021-07-24
Update gcc on AT next

Cherbot:
Bump to revision ead235f60139


mscastanho:
Again, a GCC off-tree patch needs to be remove from the build. This (and some related issued) is being handled by #2241. Closing.

PR #2264, opened 2021-07-24
Update python on AT next

Cherbot:
Bump to version 3.10.0
Bump to revision 5ec275758dbc


mscastanho:
Failed when building boost. But only on SLES15 on ppc64le:
```
./boost/python/detail/wrap_python.hpp:82:10: fatal error: patchlevel.h: No such file or directory
   82 | #include <patchlevel.h>
      |          ^~~~~~~~~~~~~~
compilation terminated.
```

PR #2260, opened 2021-07-24
Update glibc on AT next

Cherbot:
Bump to revision d34ed66f96fa

PR #2293, opened 2021-08-09
Diverge glibc on AT 15.0 from AT next

er-1:
glibc 2.34 has been released and will be distributed on AT 15.0.
AT next will continue to track glibc master branch.


er-1:
Failing on the same libhugetlbfs issue as in PR #2260

PR #2295, opened 2021-08-13
Disable use of morecore by libhugetlbfs on AT 15 and next

mscastanho:
Starting with glibc 2.32, __morecore hook has been marked as deprecated, and was
completely removed on glibc 2.34, which causes an undefined symbol error during
the build of libhugetlbfs. This affects AT 15 and current AT next.
   
Greater changes are needed in order to keep providing the same functionality
with future versions of glibc. Meanwhile, we can disable hugepage-backed heap
setup if __morecore is not available so users can at least keep using the other
features provided by the library.

I plan to submit the patch being applied here upstream, but for such it requires some
changes to the testsuite. But the current version is enough for AT's needs.

This PR includes #2293 to make sure these changes fix that issue.

Closes #2260
Closes #2293


tuliom:
This glibc commit ID is out-of-date. It's pre-glibc 2.34.
Can we use a newer one, e.g. 79474303223c5665bec75ffbdb2a86ee04a2514b

tuliom:
> This glibc commit ID is out-of-date. It's pre-glibc 2.34.
> Can we use a newer one, e.g. 79474303223c5665bec75ffbdb2a86ee04a2514b

We might also need to integrate PR #2293 in this PR.

mscastanho:
@er-1 @tuliom I replaced the old glibc update by the commits from #2293.

mscastanho:
`libhugetlbfs` build failed only on Ubuntu and Debian. Taking a look at the logs, I see:
```
<stdin>:1:1: error: expected identifier or ‘(’ before ‘-’ token
<stdin>:1:4: error: stray ‘#’ in program
```

This is being caused by the patch I added to libhugetlbfs.
But note this does not happen on other distros, and the code is exactly the same. I think I'm hitting [this `make` bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=976335) on Debian/Ubuntu.

A small reproducer:
```
$ cat /tmp/Makefile.broken 
BROKEN := $(shell echo -e '\#include <stdio.h>\nint main() { printf("Hello!"); return 0; }' | gcc -c -xc -o /dev/null - &> /dev/null)

all:
        @echo ok
```

On RHEL8 / Fedora 33 / SLES 15:
```
$ make -f /tmp/Makefile.broken 
ok
```

On Debian 10 / Ubuntu 20.04:
```
$ make -f ~/Makefile.broken 
<stdin>:1:1: error: expected identifier or ‘(’ before ‘-’ token
<stdin>:1:4: error: stray ‘#’ in program
ok
```

mscastanho:
Turns out this can be avoided if we use `/usr/bin/echo` instead of just `echo`. I updated the patch and pushed the changes 🤞🏼

tuliom (approved):

tuliom:
This PR has one approval, so I enabled auto-merge.

er-1:
Failed when building AT 15.0 on  x86_64 for debian 10 and ubuntu 20.04:
```
++ make libs tools -j1
make[1]: warning: -j1 forced in submake: resetting jobserver mode.
make[1]: Entering directory '/home/tcbot/bot-worker/at/src/at15.0-0-rc1.debian-10_x86_64_ppc64le/builds/libhugetlbfs_1'
<stdin>: In function ‘morecore_exists’:
<stdin>:2:36: error: ‘__morecore’ undeclared (first use in this function)
<stdin>:2:36: note: each undeclared identifier is reported only once for each function it appears in
	VERSION
version update: at15.0-0-rc1-4-geaf211b8
version string: at15.0-0-rc1-4-geaf211b8
<stdin>: In function ‘morecore_exists’:
<stdin>:2:36: error: ‘__morecore’ undeclared (first use in this function)
<stdin>:2:36: note: each undeclared identifier is reported only once for each function it appears in
	CC64 obj64/elflink.o
	AS64 obj64/sys-elf64lppc.o
	CC64 obj64/elf64lppc.o
	CC64 obj64/hugeutils.o
	CC64 obj64/version.o
version.c:3:19: warning: ‘libhugetlbfs_version’ defined but not used [-Wunused-const-variable=]
    3 | static const char libhugetlbfs_version[] = "VERSION: "VERSION;
      |                   ^~~~~~~~~~~~~~~~~~~~
	CC64 obj64/init.o
	CC64 obj64/morecore.o
morecore.c: In function ‘__lh_hugetlbfs_setup_morecore’:
morecore.c:375:17: error: ‘__morecore’ undeclared (first use in this function); did you mean ‘thp_morecore’?
  375 |                 __morecore = &thp_morecore;
      |                 ^~~~~~~~~~
      |                 thp_morecore
morecore.c:375:17: note: each undeclared identifier is reported only once for each function it appears in
make[1]: *** [Makefile:304: obj64/morecore.o] Error 1
```

The same builds for AT next passed.

mscastanho:
Taking a look.

The builds for next are expected to pass because they still contain an old glibc revision, before the total removal of `__morecore`, so they will just issue warnings, but won't fail. I didn't bump the version for next here to avoid further surprises that could hold this PR back.

glibc on AT 15 on the other hand is up to date in this PR.

mscastanho:
dash was being used as the shell for make on Debian and Ubuntu, and that was causing issues with the `+=` operator. I added a second patch to force make to always use `/bin/bash` instead, which should fix the issue.

PR #2296, opened 2021-08-14
Update gcc on AT next

Cherbot:
Bump to revision 261512fa6d56


tuliom:
I have a patch fixing this, but it's requiring a lot of changes to build glibc.

PR #2302, opened 2021-08-14
Update gcc on AT 13.0

Cherbot:
Bump to version 9.4.1
Bump to revision e3b2c2381655


tuliom:
I have a patch fixing this, but it's requiring a lot of changes to build glibc.

PR #2307, opened 2021-08-16
Move the definition of AT_KERNEL to base.mk on AT >= 15.0

tuliom:
Give a new interpretation to AT_KERNEL by moving it to base.mk.  Now,
it will be treated as the minimum kernel version supported by one AT
version.
    
This new interpretation will allow software built on a distro with a
newer kernel version to be executed on an older kernel version.
    
This is particularly more interesting when users start to run their
software inside containers whose host kernel may not be the same of the
distro inside the container.


er-1 (approved):

rzinsly (approved):

tuliom:
This PR already has 2 approvals, so I'm enabling auto-merge.

PR #2331, opened 2021-08-31
Verify `systemctl` is available before using it

### LINKED ISSUE #2330: Stop assuming systemd/systemctl is always available
Containers may not have systemd/systemctl installed even if they're based on Linux distributions that support only systemd.

This leads to the following kind of error:

```
/var/lib/dpkg/info/advance-toolchain-at15.0-beta1-runtime.postinst: line 30: systemctl: command not found
 Failed to start at150-0-beta1-cachemanager.service.
 AT's linker cache file is not automatically updated when a package is installed or modified.
 Please run /opt/at15.0-0-beta1/bin/watch_ldconfig if you want it to be.
 ```
 
 This is reproducible with https://github.com/advancetoolchain/at-docker

er-1:
In packages scriplets, check that `systemctl` is available before using it.
This will avoid error messages when installing AT packages.

Fix #2330

### DIFF
--- a/configs/13.0/deb/monolithic/runtime.postinst
+++ b/configs/13.0/deb/monolithic/runtime.postinst
@@ -1,6 +1,6 @@
 #!/bin/bash
 #
-# Copyright 2017 IBM Corporation
+# Copyright 2021 IBM Corporation
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


tuliom (approved):
The patch looks good to me, but I think that in order to fix issue #2330 , the same changes would be needed in the prerm/postinst files for Debian/Ubuntu.

er-1:
> The patch looks good to me, but I think that in order to fix issue #2330 , the same changes would be needed in the prerm/postinst files for Debian/Ubuntu.

My bad, I missed the Debian/Ubuntu files.
That's fixed.

tuliom (approved):

mscastanho (approved):

PR #2350, opened 2021-09-08
Remove binutils/stage_3

ThinkOpenly:
`binutils/stage_3` has been obsolete since commit
8c3afc9b1250338a026ac55957c0416829d4c02d, and should've been
removed then.  Remove it now.


er-1 (approved):

rzinsly (changes requested):
This will leave broken links on AT 15.0 and next, these files need to be removed as well
Also, it would be better to mention that this change affects AT >= 14.0 in the title or in the commit message.

ThinkOpenly:
> This will leave broken links on AT 15.0 and next, these files need to be removed as well

Oops! You are correct of course, not sure how I missed those.  Fixed in a force push just now.

> Also, it would be better to mention that this change affects AT >= 14.0 in the title or in the commit message.

Added that exact text in the commit subject.  Thanks for the review!

rzinsly (approved):
LGTM!

PR #2353, opened 2021-09-11
Update binutils on AT next

Cherbot:
Bump to revision 5abdee61d997


rzinsly:
The AT next RHEL8 ppc64le failed in the test_ssl python test due to a timeout, I built it manually and passed. The server was just starting and it is possible that it had more things running during our CI build than mine. Anyway this is not related to this patch as it failed on others as well, so I'm merging it, we can create a issue if this happens again.

PR #2355, opened 2021-09-11
Update glibc on AT next

Cherbot:
Bump to revision abd383584b16


rzinsly:
Failed with the same issue as #2353, merging it.

PR #2362, opened 2021-09-11
Update python on AT 15.0

Cherbot:
Bump to revision 1b02e29fb275


rzinsly:
AT 15.0 RHEL8 ppc64le failed with a timeout on the `test_subprocess` Python test, it could be something similar to what happened with #2353.

rzinsly:
I was able to build it manually, this should be related to #2368, merging it.

PR #2361, opened 2021-09-11
Update glibc on AT 15.0

Cherbot:
Bump to revision 52d011974318

PR #2366, opened 2021-09-12
Fix builds on RHEL 8

ThinkOpenly:


rzinsly (approved):
>One option is to backport those same out-of-tree patches to AT,
but that adds a support burden in a security-sensitive area.

Another option is to disable `debuginfod` entirely on all builds.
For now, this patch LGTM.

rzinsly:
AT 14.0 RHEL8 ppc64le failed with the same error as #2361, doesn't seem related to the changes of this patches.

tuliom (approved):

Issue #2352, opened 2021-09-10
Important issues in Binutils and GCC for AT 15.0-1


tuliom:
We'd like to use this issue in order to track fixes for Binutils and GCC for AT 15.0-1.

- [x] [Binutils bug 28192](https://sourceware.org/bugzilla/show_bug.cgi?id=28192)
- [x] https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=b0963c4379becafaebd8e52b0b42698ff151c293
- [x] https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=69feb7601e86274fa9abbfb420b00c8adf947e7b
- [x] https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=6278065af07634278ba30029d92a82b089969baa
- [x] https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=42a9e4e1381ba14d8ef21b331494945c2f51e6be

peter-bergner:
@tuliom: what form do you want this information?  I have some IBM GCC 11 git log commits for those entries I think are important.  Do you just want me to cut/paste that here or just give a synopsis of the entries?

tuliom:
> what form do you want this information?

@peter-bergner Whatever is best for you.  Commit IDs are great! Bugzilla entries too.

> Do you just want me to cut/paste that here or just give a synopsis of the entries?

Whatever is easier for you.  I just want to track if the important fixes have properly landed in AT 15.0-1.

peter-bergner:
It's easiest just to cut/paste git log entries, so if you're fine with that, then that's what I'll do.

peter-bergner:
Some of the entries are pretty large, so I think I will remove the ChangeLog portion of the commit logs and just give you the commit log message minus the ChangeLog entry.

peter-bergner:
```
commit da0a5628d47ee59320dac9787f3232b0911fbc14
Author:     Segher Boessenkool <segher@kernel.crashing.org>
AuthorDate: Wed Sep 8 13:10:30 2021 +0000
Commit:     Segher Boessenkool <segher@kernel.crashing.org>
CommitDate: Tue Sep 21 23:22:44 2021 +0000

    rs6000: Fix ELFv2 r12 use in epilogue
    
    We cannot use r12 here, it is already in use as the GEP (for sibling
    calls).
    
    (cherry picked from commit 86e6268cff328e27ee6f90e2afc35b6f437a25cd)

[...]

PR #2393, opened 2021-09-29
Update the list of patches for Binutils on AT 15.0

tuliom:
- Remove an unnecessary patch.
- Add patches for issue #2352 .


mscastanho (approved):
LGTM.

I'd prefer if we could backport these to binutils 2.37 upstream, but I'm ok with having them applied off-tree meanwhile so we can get the bug fix.

PR #2395, opened 2021-10-02
Update gcc on AT next

Cherbot:
Bump to revision 5f0285418940


er-1:
It failed while compiling binutils:
```
/home/tcbot/bot-worker/at/src/at16.0-0-alpha.debian-10_ppc64le_ppc64le/sources/binutils/binutils/readelf.c: In function ‘find_section’:
/home/tcbot/bot-worker/at/src/at16.0-0-alpha.debian-10_ppc64le_ppc64le/sources/binutils/binutils/readelf.c:349:8: error: the comparison will always evaluate as ‘true’ for the pointer operand in ‘filedata->section_headers + (sizetype)((long unsigned int)i * 80)’ must not be NULL [-Werror=address]
  349 |   ((X) != NULL                                                          \
      |        ^~
```

mscastanho:
This is caused by a new (or just improved?) warning on GCC. Binutils may require a small change like [this one](https://sourceware.org/git/?p=glibc.git;a=commit;h=a312e8fe6d89f5eae6a4583d5db577121e61c0b5) from glibc.

PR #2400, opened 2021-10-02
Update valgrind on AT next

Cherbot:
Bump to revision 345edda43951


er-1:
It failed while applying a patch to Valgrind source code.

PR #2401, opened 2021-10-02
Update binutils on AT 15.0

Cherbot:
Bump to revision 2e62356d25e9


er-1:
It failed while applying a patch to binutils source code.

PR #2440, opened 2021-10-18
Update mpfr repository

er-1:
INRIA has moved mpfr source code from a SVN repository to a GIT repository.


ThinkOpenly:
> Doing a quick grep on the repo looks like this was the last SVN project we used. We may be able to remove the build requirement for SVN in the near future.

That would be awesome, because it seems that very recent versions of SVN exhibit issues on very recent versions of, at least, Red Hat (RH bugzilla 2004714).

PR #2443, opened 2021-10-20
Support building on CentOS Stream 9

ThinkOpenly:
A few assorted changes:
- I cannot find the package which, on RHEL8, contained `lsb_release`,
  so I copied a version from RHEL8 into `scripts/helpers` and, in
  `Makefile`, fall back to it if `lsb_release` cannot be found.
- Created a new `configs/next/distros/redhat-9.mk` file, based on the
  very similar `redhat-8.mk` file. Notable changes:
  - Packages which could not be found: autogen-libopts, libstdc++-static
    redhat-lsb-core (contained `lsb_release`), imake.
  - Packages which are versioned in the package name and need to be
    updated: python36 became python3 (3.10).
  - Packages which are no longer needed: subversion.  (This is possible
    due to commit 4689341d939c7dc9e16026912cb2b342f3c9670b.)


tuliom:
> I cannot find the package which, on RHEL8, contained lsb_release

It seems the package is now called [redhat-lsb](https://gitlab.com/redhat/centos-stream/rpms/redhat-lsb).
It looks like they haven't built it yet: https://composes.stream.centos.org/test/latest-CentOS-Stream/compose/BaseOS/source/tree/Packages/

tuliom (approved):

PR #2427, opened 2021-10-16
Update gcc on AT next

Cherbot:
Bump to revision 93d183a5fff9

PR #2480, opened 2021-11-09
Update gcc, glibc and boost on AT next

rff:
Boost 1.76.0 on AT next is incompatible with gcc 12/trunk, but there is now boost 1.77.0 that is compatible.
So before updating gcc we need to update boost to 1.77.0. Gcc trunk also found a bug on glibc that was just fixed on master, so we need to update glibc before updating gcc.

Each commit was tested in sucession to make sure there is no intermediate broken state.

This update supersedes PR #2427 that was bumping gcc revision.


mscastanho (approved):
LGTM.

When we need to break a symlink (like you did for `configs/next/packages/boost/sources`) we usually copy the file in one commit and leave any additions/modifications to a separate commit, to make it clear what changed. 

But no need to modify the PR and test everything again, I'm ok with it as-is. Thanks!

rff:
> When we need to break a symlink (like you did for `configs/next/packages/boost/sources`) we usually copy the file in one commit and leave any additions/modifications to a separate commit, to make it clear what changed.
> 
> But no need to modify the PR and test everything again, I'm ok with it as-is. Thanks!

But now I can't stop thinking about it :D
Just updated the PR with the better commit separation.

mscastanho (approved):

rzinsly (approved):

PR #2496, opened 2021-11-16
Update gcc on AT next and stop applying patch

mscastanho:
Bump to revision 2f3d43a35155.

The patch to remove the forced runpath when using --with-advance-toolchain has
been merged upstream as 9598134a055e9426fa283e51dac178d11af2c668, so we don't
need to apply it off-tree anymore.

Closes #2483


er-1 (approved):

mscastanho:
There was a segfault on `gcc_4`. This seems to be fixed with the latest GCC revision, so updated the PR accordingly.

PR #2497, opened 2021-11-16
Fix expat source link and version in AT 15.0

rff:
Expat source file from AT 15.0 was pointing to AT next source instead of the other way around when the version is already released.

This patch broke this link make both independent.  Since expat sources was updated two times in AT next since the last release from AT 15.0 (at15.0-1), the version copied to AT 15.0 is the one from the last release and not the current version from AT next.

I don't know if expat should be kept updating after at15.0-1 to the same versions that was updated to AT next, if that is the case then I have to change this PR to take the latest version of expat/sources and maybe can even link AT next to the AT 15.0 version. But I am not sure.

### DIFF
--- a/configs/15.0/packages/expat/sources
+++ b/configs/15.0/packages/expat/sources
@@ -1,1 +1,43 @@
-../../../next/packages/expat/sources
+#!/usr/bin/env bash
+#
+# Copyright 2017 IBM Corporation
+#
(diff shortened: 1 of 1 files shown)


rff:
After reviewing commit 3bae68053b51: "Let expat from AT 15.0 track expat from AT next".
It is indeed the expected behavior until there is a new release. So I am closing this PR.

PR #2560, opened 2021-12-15
Remove old configs

tuliom:
Remove old config sets that we can't test automatically anymore.
Leave just 12.0 or newer.


er-1 (approved):

mscastanho (approved):
Nice change!

PR #2561, opened 2021-12-17
Remove references to old versions of AT

er-1:
Configsets of AT < 12 were removed in https://github.com/advancetoolchain/advance-toolchain/pull/2560
Removed references to these old version of AT.


mscastanho (commented):
Looking very good so far. Thank you for working on this.

fvtr/speccpu2006/speccpu2006.exp
```diff
@@ -61,20 +61,7 @@ if { $TARGET64 } {
...
-		set cfg_set ""
-	}
-	"^7.0|^7.1|^8.0|^9.0" {
-		set cfg_set "7.0"
-	}
-	default {
-		set cfg_set "10.0"
-	}
-}
+set cfg_set "12.0"
```
tuliom, line 64: We might consider to drop this test in favor of speccpu2017.

tuliom (approved):
Nice simplification!

fvtr/speccpu2006/speccpu2006.exp
tuliom: I don't think this needs to be done for the PR, though.

tuliom:
IMO, this patch is ready to be merged.
All tests passed.  The only failure that appeared was unrelated to this PR.
The only question that remains open is the Draft status it has.
Does that mean we should not merge it now?

PR #2582, opened 2021-12-25
Update libnxz on AT next

Cherbot:
Bump to version v0.61
Bump to revision ccfa3347ff74


tuliom:
We have to change how libnxz installs itself. There is another change we could adopt too.
I'm preparing a patch.

PR #2589, opened 2022-01-01
Update binutils on AT next

Cherbot:
Bump to revision f01e6568a892


mscastanho:
binutils_2 failed on the install step. The only error I can find in the logs is this:
```
/bin/install: cannot create regular file '/home/tcbot/bot-worker/[...]/at_opt/at-next-16.0-0-alpha/share/info/as.info': File exists
```
Which is a bit unexpected, especially because the other cross builds did not fail. The commits included in this update also don't point to anything install-specific, so I'm not sure what may have happened. I'll restart this build.

mscastanho:
Actually, it's already running. So let's wait and see.

tuliom:
> Actually, it's already running. So let's wait and see.

@mscastanho Sorry, I did that but I forgot to submit the comment I had written.
I agree with you: this failure is unexpected.  Especially because it passed in all other distros.

tuliom (approved):

tuliom:
The new test passed as expected.
I'm merging this now.

PR #2604, opened 2022-01-08
Update libnxz on AT next

Cherbot:
Bump to version v0.61
Bump to revision f627199e125c

PR #2617, opened 2022-01-10
Get the correct libnxz package version automatically

mscastanho:
update_revision.sh is still setting libnxz version to v0.61, even though v0.62
has been out for almost 6 months. Since the project now uses autoconf, we can
parse the AC_INIT string in configure.ac to properly get the latest version.


mscastanho:
See for example #2604, which is (erroneously) bumping libnxz version to v0.61

rzinsly (approved):

er-1 (approved):

PR #2618, opened 2022-01-11
Update libnxz on AT next

Cherbot:
Bump to version 0.62
Bump to revision e1cdc4646378


mscastanho:
I just noticed the new version doesn't have the `v` prefix. The git tags in libnxz repo use the `v` prefix, but since Cherbot is getting the version info directly from `configure.ac`, it just gets the bare version number.

@er-1 Do you foresee something that could break because of that?

er-1:
> @er-1 Do you foresee something that could break because of that?

No. The build and the tests passed, I think it's ok to merge.

PR #2630, opened 2022-01-15
Update glibc on AT 14.0

Cherbot:
Bump to revision bd407fd37421


er-1:
The failure isn't related to the patch. It's a network issue: "Could not resolve host: github.com"

PR #2631, opened 2022-01-15
Update openssl on AT 13.0

Cherbot:
Bump to revision 4c5c2a5efbc3


er-1:
The failure isn't related to the patch. It's a network issue: "Could not resolve host: github.com"

PR #2634, opened 2022-01-18
Move CPU-optimized libraries under `glibc-hwcaps`

ThinkOpenly:


mscastanho:
@ThinkOpenly the builds failed in the FVTR step:
```
ck_tuned		power9: 	failed
```
You'll also need to change that test to guarantee it looks for tuned libraries in the new directory. But remember that file is shared with other AT versions, so you may have to add a version guard as well.

ThinkOpenly:
Let me resubmit this as a new PR...

PR #2640, opened 2022-01-22
Update gcc on AT next

Cherbot:
Bump to version 12.0.1
Bump to revision abe3a4f0e9c4


er-1:
Failing while building `glibc` on stage 1:
```
localealias.c: In function ‘read_alias_file’:
localealias.c:335:56: error: pointer may be used after ‘realloc’ [-Werror=use-after-free]
  335 |                               map[i].alias += new_pool - string_space;
      |                                               ~~~~~~~~~^~~~~~~~~~~~~~
localealias.c:325:49: note: call to ‘realloc’ here
  325 |                       char *new_pool = (char *) realloc (string_space, new_size);
      |                                                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

mscastanho:
@er-1 These were caused by the new `-Wuse-after-free` warning added to GCC 12. This specific instance on glibc is already known to GCC/glibc developers and is apparently benign (bugs [28521](https://sourceware.org/bugzilla/show_bug.cgi?id=28521) and [26779](https://sourceware.org/bugzilla/show_bug.cgi?id=26779)). We should add `-Wno-error=use-after-free` to (probably) all glibc stages for the time being and create an issue so we remember to remove it in the future.

mscastanho:
Now `gcc_power10` failed with an ICE. I'm testing with the latest GCC revision to see if that's gone.

PR #2652, opened 2022-01-24
Update gcc and glibc on AT next

mscastanho:
Bump to version 12.0.1
Bump to revision abe3a4f0e9c4

This new batch of changes includes a new use-after-free warning to GCC which
causes issues when building glibc (glibc bugs 28521 and 26779). Let's disable
these warnings for now until they are fixed upstream.

Closes #2640


mscastanho:
The glibc warning has already been fixed, so no need to disable it anymore. But I bumped into an ICE when updating GCC. It's being taken care of by the GCC folks, so I'll wait a bit more until the fix lands upstream to get the latest revisions and update both GCC and glibc at once.

mscastanho:
This is the bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=104335

mscastanho:
The GCC bug has been fixed, and glibc already has a fix for the warning, so this is now just updating both to more recent (but not latest) revisions. I'll let Cherbot get the latest and greatest on Saturday, to avoid further surprises until then.

mscastanho:
My bad, I accidentally used the wrong GCC commit (without the fix). I'm testing a build with the correct one and will update the PR once it finishes.

mscastanho:
This is now failing on `gcc_4`:
```
/home/tcbot/bot-worker/at/src/at16.0-0-alpha.suse-15_ppc64le_ppc64le/sources/gcc/libgomp/openacc.f90:73:7:

   73 |   use openacc_kinds
      |       1
Fatal Error: File ‘openacc_kinds.mod’ opened at (1) is not a GNU Fortran module file
compilation terminated.
```
I'm investigating.

tuliom:
@mscastanho I'm bisecting the Fortran issue.
So far, I've seen that commit e97cfaa9f614 + a patch from fac15bf84807 works well.

While this is not a new enough commit version, it will help to update GCC a little bit.

mscastanho:
@tuliom found out the GCC commit causing the issue was `3f30f2d1dbb rs6000: Fix LE code gen for vec_cnt[lt]z_lsbb [PR95082]`, which altered the semantics of a builtin used by one of our zlib optimizations. I updated the zlib patch with a fix.

rzinsly (approved):

rff (approved):

PR #2685, opened 2022-02-12
Update glibc on AT 15.0

Cherbot:
Bump to revision 007e054d786b


er-1:
The build failure on SLES 15 is because a machine issue (lack of disk space). I think it's safe to merge.

PR #2686, opened 2022-02-12
Update python on AT 15.0

Cherbot:
Bump to revision a7c1cc416967


er-1:
The build failure on SLES 15 is because a machine issue (lack of disk space). I think it's safe to merge.

PR #2691, opened 2022-02-19
Update gcc on AT next

Cherbot:
Bump to version 12.0.1
Bump to revision 5a9ba3f27f3a


mscastanho:
Failed on glibc_1:
```
localealias.c: In function ‘read_alias_file’:
localealias.c:335:56: error: pointer may be used after ‘realloc’ [-Werror=use-after-free]
  335 |                               map[i].alias += new_pool - string_space;
      |                                               ~~~~~~~~~^~~~~~~~~~~~~~
localealias.c:325:49: note: call to ‘realloc’ here
  325 |                       char *new_pool = (char *) realloc (string_space, new_size);
      |                                                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

This has already been fixed upstream on glibc. But we need to bump glibc and gcc together to get all fixes. This will be done in #2652

PR #2730, opened 2022-03-08
Add support of Ubuntu 22.04 on AT next

er-1:


tuliom:
There is something wrong with the Python build:
```
$ ~/opt/at-next-16.0-0-alpha/bin/python3 -c "import sys; print(sys.prefix)"
/usr
$ ~/opt/at-next-16.0-0-alpha/bin/python3 -c "from sysconfig import get_paths as gp; print(gp()['include'])"
/usr/local/include/python3.10
```

er-1:
> There is something wrong with the Python build

Any idea how to fix that?

tuliom:
> Any idea how to fix that?

Not yet, but I've just found out that `/lib/powerpc64le-linux-gnu/libpython3.10.so.1.0` is being used instead of the newly built library.
We're missing an `ldconfig` between `python_1` and `boost_1 gdb_1`.

er-1:
After an update of the Ubuntu 22.04 machine, it passed. Not sure if it's a good thing.

tuliom:
> After an update of the Ubuntu 22.04 machine, it passed. Not sure if it's a good thing.

I suspect it's just hiding the issue.

er-1:
I'm going to merge this PR.
An issue will be opened to keep track of the python issue.

PR #2731, opened 2022-03-09
Update mpfr on AT >= 12.0

er-1:
Get sources from the tarball instead of the git repository.
This avoid autoconf issues with the latest version of autoconf.


er-1:
The failure on Ubuntu 22.04 is expected as AT doesn't support Ubuntu 22.04 yet.

PR #2733, opened 2022-03-12
Update expat on AT next

Cherbot:
Bump to revision bccf36869d4f


er-1:

PR #2744, opened 2022-03-18
Ignore test_tarfile Python test for older releases

mscastanho:
This test is known to fail if run on tmpfs on ppc64 [1]. It was fixed
for Python 3.8, which maps to AT 14, so it will fail for any versions
prior to that when tmpfs is used (like inside a container).

[1] https://bugs.python.org/issue35772


er-1 (approved):

rzinsly (approved):

mscastanho:
The Ubuntu 22 build failure is unrelated to this PR. Support is being added by #2730.

PR #2748, opened 2022-03-19
Update glibc on AT next

Cherbot:
Bump to revision 59c30d5708e4


mscastanho:
The issue with Ubuntu22 is unrelated to glibc and is being handled by #2730

Issue #2557, opened 2021-12-13
dnf does not support failovermethod anymore


tuliom:
dnf stopped accepting a failovermethod in yum repositories configurations and started to warn about it.
I can reproduce this on RHEL 8.5.

We need to remove this option from our documentation, e.g. https://www.ibm.com/support/pages/advtool-install#redhattab2

Reference: https://bugzilla.redhat.com/show_bug.cgi?id=1653831

PR #2755, opened 2022-03-25
removing failovermethod option from documentation

fabiane-watanabe:
Removing failovermethod option from documentation.

### DIFF
--- a/configs/13.0/release_notes/release_notes-inst.html
+++ b/configs/13.0/release_notes/release_notes-inst.html
@@ -135,7 +135,6 @@ endif_RHEL7 -->
 [__VERSION__]
 name=Advance Toolchain IBM FTP
 baseurl=ftp://public.dhe.ibm.com/software/server/POWER/Linux/toolchain/at/redhat/RHEL8
-failovermethod=priority
 enabled=1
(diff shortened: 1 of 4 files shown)


mscastanho (changes requested):
There are other mentions to `failovermethod=priority` on `configs/14.0/release_notes/release_notes-inst.html`. Example:
https://github.com/advancetoolchain/advance-toolchain/blob/1233554a7120a6f1a9e057649e0ddb4e34d0effe/configs/14.0/release_notes/release_notes-inst.html#L77

Looks like this file is used to build the Install instructions in the release notes, so we should remove the occurrences on that file as well. Note that the corresponding files for AT 15 and AT next are symlinks that end up pointing to this one, so changing this one will change for all three versions at once, which I think is OK.

Also it would be nice to mention #2557 either in the commit message or the PR body, since it's fixing that issue.

rzinsly (approved):
LGTM

er-1 (approved):

mscastanho:
This PR is ready to go, it had passed all tests, and the changes added by the rebase are unrelated to this PR, so no need to run the tests again.

PR #2840, opened 2022-05-03
Clone AT next into AT 16.0

er-1:
Prepare AT 16.0 based on AT next for further changes.


configs/next/deb/monolithic/focal_dh_change_dest.patch
```diff
@@ -1 +1 @@
-../../../15.0/deb/monolithic/focal_dh_change_dest.patch
\ No newline at end of file
+../../../16.0/deb/monolithic/focal_dh_change_dest.patch
```
mscastanho, line 1: Are we still going to support Ubuntu Focal for AT 16? If not, I suspect we don't need this file.

configs/next/deb/monolithic/buster_dh_change_dest.patch
```diff
@@ -1 +1 @@
-../../../15.0/deb/monolithic/buster_dh_change_dest.patch
\ No newline at end of file
+../../../16.0/deb/monolithic/buster_dh_change_dest.patch
```
mscastanho, line 1: Are we still going to support Debian Buster for AT 16? If not, I suspect we don't need this file.

configs/next/deb/monolithic/focal_dh_change_dest.patch
er-1: Good catch. I forgot to remove it.

configs/next/deb/monolithic/buster_dh_change_dest.patch
er-1: I forgot this one too.

er-1:
`configs/next/packages/gcc/sources` isn't a symbolic link to `configs/16.0/packages/gcc/sources` because of #2841 that updated the gcc version to 13.0.
gcc 12.0 is tracked by AT 16.0.
gcc 13.0 is tracked by AT next.

rff (approved):

mscastanho (approved):

PR #2862, opened 2022-05-16
Update libhugetlbfs patch to remove morecore

mscastanho:
The patch to disable morecore usage on libhugetlbfs is causing problems
when building AT 16 on Debian 11 x86 because it uses /usr/bin/echo
instead of /bin/echo. Use the version submitted upstream instead (merge
pending), which does not have this issue.


er-1:
About RHEL9 failure:
- I added the kernel version for RHEL to the cache.
- it failed to build `openssl`
- there is something wrong with the logs

mscastanho:
@er-1 openssl failed because the RHEL9 server was missing `perl-FindBin` (installed now). We should add it here to make sure the build fails earlier.

https://github.com/advancetoolchain/advance-toolchain/blob/b7f7be35ef39af13ae3018e7b5f8ab5482ac9186/configs/16.0/distros/redhat-9.mk?plain=1#L67-L82

The logs were not uploaded because the server was missing the proper keys. Should be fixed now.

rzinsly (approved):

Issue #2888, opened 2022-06-02
Failing of package installation in at-docker container


fabiane-watanabe:
When the package to be installed has shared libraries dependencies,  installation fails because AT loader doesn’t find these libraries ( that has just installed ).

Steps to reproduce ( Debian 10/ubuntu 20 - at15): 
- install  systemd using apt-get install    
        apt-get install system d 

Some systemd libraries dependencies are installed as well, but when  systemd-machine-id-setup command is executed, a fail occurs:
“
      find library=libcryptsetup.so.12 [0]; searching
       search cache=/opt/at15.0/etc/ld.so.cache
       search path=/opt/at15.0/lib64/………………………..	(system search path)

      
systemd-machine-id-setup: error while loading shared libraries: libcryptsetup.so.12: cannot open shared object file: No such file or directory
dpkg: error processing package systemd (--configure):
 installed systemd package post-installation script subprocess returned error exit status 127   “

PR #2889, opened 2022-06-03
Fix installation fail of some packages in at-docker

fabiane-watanabe:
Compile glibc with “user-defined-trusted-dirs” option to add
standard system libraries to paths that dynamic loader searchs for.
These paths will be the last in the order that dynamic loader will
look for( after  AT libraries paths )

Fixes issue #2888

### DIFF
--- a/configs/15.0/packages/glibc/stage_2
+++ b/configs/15.0/packages/glibc/stage_2
@@ -118,7 +118,14 @@ atcfg_configure() {
 
 # Make build command
 atcfg_make() {
-	PATH=${at_dest}/bin:${PATH} ${SUB_MAKE}
+	
+	if [[ ("${distro_fm}" == "ubuntu") || ("${distro_fm}" == "debian") ]]; then
+			user_defined_dir='user-defined-trusted-dirs=/lib/powerpc64le-linux-gnu:/usr/lib/powerpc64le-linux-gnu'			
+	else
+			user_defined_dir='user-defined-trusted-dirs=/lib64:/usr/lib64'          
+	fi
+
+	PATH=${at_dest}/bin:${PATH} ${SUB_MAKE} ${user_defined_dir}			
 }
 
 # Conditional install build command


configs/15.0/packages/glibc/stage_2
```diff
@@ -118,7 +118,19 @@ atcfg_configure() {
 
 # Make build command
 atcfg_make() {
-	PATH=${at_dest}/bin:${PATH} ${SUB_MAKE}
+	if [[ ("${distro_fm}" == "ubuntu") || ("${distro_fm}" == "debian") ]]; then
+			if [[ "${build_arch}" == "ppc64le" ]]; then
+					user_defined_dir='user-defined-trusted-dirs=/lib/powerpc64le-linux-gnu:/usr/lib/powerpc64le-linux-gnu'
+			elif [[ "${build_arch}" == "ppc64" ]]; then
+					user_defined_dir='user-defined-trusted-dirs=/lib/powerpc64-linux-gnu:/usr/lib/powerpc64le-linux-gnu'
```
mscastanho, line 125: There is a typo in this line: should be `/usr/lib/powerpc64-linux-gnu` instead of `/usr/lib/powerpc64le-linux-gnu`.

But AT does not support ppc64 (BE) anymore, so I don't think there's a need to add it here. It would be even hard to test it properly.

mscastanho (changes requested):

configs/15.0/packages/glibc/stage_2
```diff
@@ -118,7 +118,19 @@ atcfg_configure() {
...
+			else    
+					user_defined_dir='user-defined-trusted-dirs=/lib/x86_64-linux-gnu:/usr/lib/x86_64-linux-gnu'
```
tuliom, line 127: I think this is wrong, because we don't build AT for x86_64.
Is the code ever reaching this part?

tuliom (changes requested):

mscastanho (approved):

tuliom (approved):

PR #2901, opened 2022-06-10
Improve fvtr.sh script

er-1:
`AT_BASE` is needed by some FVTR scripts, but `AT_BASE` could point to a directory
that doesn't exist. This happens when FVTR is run against installed AT packages,
with a different working directory than the one used for the build.

This patch updates `AT_BASE` if its current value doesn't exist. The new value
is based on the path of the running `fvtr.sh` script.


fvtr/fvtr.sh
```diff
@@ -167,6 +167,11 @@ then
 	export AT_WD=$AT_WD
 fi
 
+if [ ! -e "${AT_BASE}" ]
+then
+	export AT_BASE=$(realpath "$(dirname $0)/..")
+fi
```
mscastanho, line 173: Is it possible that `AT_BASE` dir exists but is still not valid for this script? Perhaps we could drop the if and always set it here instead of in the Makefile as done in e1a7ebe107b0989db.

er-1: If `AT_BASE` dir exists and is not valid, then I think the tests will fail.
I agree to make a change in the Makefile.

tuliom:
While investigating another issue, I unintentionally caused `AT Pull Request Build 797` to fail.
The builds should continue to run, though.

I also restarted 2 builds that had failed for reasons that appear to be unrelated to this PR.

tuliom (approved):
All tests passed now.
LGTM and ready to be merged.

PR #2903, opened 2022-06-11
Update gcc on AT next

Cherbot:
Bump to revision ef1e4d80dd30


er-1:
It failed while building `boost`:
```
./boost/intrusive/pack_options.hpp:164:29: internal compiler error: in iterative_hash_template_arg, at cp/pt.cc:1819
  164 |    static const std::size_t last_idx = sizeof_typelist<Typelist>::value - 1;
      |                             ^~~~~~~~
```

mscastanho:
I reported the bug upstream: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105956

It doesn't look like a target-specific issue, as I was able to reproduce it on x86_64 as well.

Issue #2827, opened 2022-04-29
Error using libcrypt.so on AT15.0 in Ubuntu-focal


fabiane-watanabe:
In the latest release of Ubuntu e Fedora, libcrypt.so is provided by "libxcrypt" project, instead of Glibc.

How to reproduce:
Ubuntu : 20.04
AT :15.0

tuliom:
More details: distros have been configuring glibc with `--disable-crypt`, e.g. https://launchpadlibrarian.net/596537387/buildlog_ubuntu-focal-ppc64el.glibc_2.31-0ubuntu9.9_BUILDING.txt.gz

I suspect this could cause issues to programs built with AT 15.0.  Need to confirm if libxcrypt is ABI-compatible with glibc 2.34, before changing AT 15.0.

AT next should definitely be modified though.

tuliom:
> I suspect this could cause issues to programs built with AT 15.0. Need to confirm if libxcrypt is ABI-compatible with glibc 2.34, before changing AT 15.0.

According to libxcrypt's documentation:

> On Linux-based systems, by default libxcrypt will be binary backward compatible with the libcrypt.so.1 shipped as part of the GNU C Library. This means that all existing binary executables linked against glibc’s libcrypt should work unmodified with this library’s libcrypt.so.1.

While this does not answer the previous question, it does look promising.

PR #2917, opened 2022-06-15
  Stop building  glibc´s libcrypt on AT.

fabiane-watanabe:
Since distributions (starting from RHEL8, SLES15, Ubuntu20, Debian11)
stopped building glibc´s libcrypt, it´s safer  AT stops too.

Fixes usses #2827


mscastanho:
The x86_64 builds failed on `gcc_4` because they couldn't find `crypt.h`:
```
checking sys/epoll.h presence... /home/tcbot/bot-worker/at/src/workdir/sources/gcc/libsanitizer/sanitizer_common/sanitizer_platform_limits_posix.cpp:142:10: fatal error: crypt.h: No such file or directory
  142 | #include <crypt.h>
      |          ^~~~~~~~~
compilation terminated.
make[5]: *** [Makefile:611: sanitizer_platform_limits_posix.lo] Error 1
make[5]: *** Waiting for unfinished jobs....
```

configs/14.0/packages/glibc/stage_1
```diff
@@ -140,6 +140,15 @@ atcfg_pre_configure()
...
 {
 	local base_target=$(find_build_target ${AT_BIT_SIZE})
+
+	#  In case of cross build, glibc's libcrypt will be  compiled
+	# because it´s required to build gcc.
+	if [[ ("${distro_fm}" == "redhat") ||
+		("${distro_fm}" == "suse" ) ||
+		("${DISTRO_ID}" == "ubuntu-20") ]]; then
+		disable_crypt=yes
+	fi
```
tuliom, line 150: The comment and the values being tested do not match.
Is the comment out-of-date?

configs/14.0/packages/glibc/stage_2
```diff
@@ -83,6 +83,11 @@ atcfg_pre_configure() {
 atcfg_configure() {
 	local base_target=$(find_build_target ${AT_BIT_SIZE})
 
+	if [[ ("${distro_fm}" == "redhat") ||
+		("${distro_fm}" == "suse" )  ||
+		("${DISTRO_ID}" == "ubuntu-20") ]]; then
+		disable_crypt=yes
+	fi
```
tuliom, line 90: This part doesn't have a comment.
Why is this needed?
Should this actually test if `/usr/include/crypt.h` defines `XCRYPT_VERSION_STR` ? That would say if the libxcrypt development files are available in the OS.
But be careful because the test I mentioned is only valid for native builds.  Cross builds should not test it.

tuliom (changes requested):
I think there is a part of the code that is testing the wrong thing and might have a wrong source code comment. I'm suggesting a way for testing if libxcrypt is available that is distro-agnostic.
This code is duplicated in other places.  I avoided commenting all the places.

Besides that I expected that `AT_NATIVE_PKGS_REQ` would now have `libxcrypt-devel` too (or whatever each distro calls it).

configs/14.0/packages/glibc/stage_2
fabiane-watanabe: This test doesn't work in case of Rhel8. 
In this release, libxcrypt´s version is 4.11,and  `/usr/include/crypt.h` doesn't have `XCRYPT_VERSION_STR` definition.

scripts/utilities/check_glibc_disable_crypt.sh
```diff
@@ -0,0 +1,40 @@
...
+# limitations under the License.
+#
+
+
+# Check if it´s needed to compile glibc´s libcrypt.
+# If libxcrypt is available in OS, glibc´s libcrypt will not be compiled.
+# Except in cross build. In this case, glibc's libcrypt must be  compiled because it´s required to build gcc
+
+if [[ "${cross_build}" == "no" ]]; then
+        if [[ ("${distro_fm}" == "ubuntu") || ("${distro_fm}" == "debian") ]]; then
```
mscastanho, line 24: @fabiane-watanabe I don't think this will work. Since this is a separate script, it will be called in a separate shell, so variables `cross_build` and `distro_fm` won't be defined here. You could pass these two values as arguments to this script instead.

mscastanho (changes requested):

scripts/utilities/check_glibc_disable_crypt.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+#
+# Copyright 2017 IBM Corporation
```
rzinsly, line 3: 2022

scripts/utilities/check_glibc_disable_crypt.sh
```diff
@@ -0,0 +1,40 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+# Check if it´s needed to compile glibc´s libcrypt.
+# If libxcrypt is available in OS, glibc´s libcrypt will not be compiled.
+# Except in cross build. In this case, glibc's libcrypt must be  compiled because it´s required to build gcc
```
rzinsly, line 21: This comment should be broken in two lines.

rzinsly (changes requested):
Besides changes pointed by others I have some nit style comments.

scripts/utilities/check_glibc_disable_crypt.sh
fabiane-watanabe: @mscastanho These values are environment variables ( exported by build_setenv function before this script be executed ). Also I made some tests to check if values were defined in this script.

tuliom (changes requested):
Looks mostly good to me, but I still miss this part:

> Besides that I expected that AT_NATIVE_PKGS_REQ would now have libxcrypt-devel too (or whatever each distro calls it).

scripts/utilities/check_glibc_disable_crypt.sh
fabiane-watanabe: changes made.

fabiane-watanabe:
@tuliom Is `libxcrypt-devel` still required ? Since, `crypt.h `no longer will be checked to verify `XCRYPT_VERSION_STR`

tuliom:
> @tuliom Is `libxcrypt-devel` still required ? Since, `crypt.h `no longer will be checked to verify `XCRYPT_VERSION_STR`

@fabiane-watanabe , AFAIU, other packages built by AT will link against libxcrypt.

scripts/utilities/check_glibc_disable_crypt.sh
```diff
@@ -0,0 +1,41 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# Check if it´s needed to compile glibc´s libcrypt.
+# If libxcrypt is available in OS, glibc´s libcrypt will not be compiled.
+# Except in cross build. In this case, glibc's libcrypt must be  compiled
```
rzinsly, line 20: nit: there are two spaces between **be** and **compiled**.

scripts/utilities/check_glibc_disable_crypt.sh
```diff
@@ -0,0 +1,41 @@
...
+#because it´s required to build gcc
```
rzinsly, line 21: nit: space after `#`.

rzinsly (approved):
LGTM, but there are two small changes that won't need another approval.

tuliom:
@fabiane-watanabe I'm afraid your last update brought 39 new patches unintentionally. Could you review, that, please?

tuliom:
Both AT 12.0 builds failed while building golang.
AFAICS, this is unrelated to the changes proposed here. I'm trying to reproduce the same issue without the changes in this PR.

tuliom:
I have confirmed that both AT 12.0 issues are unrelated to this PR and can be ignored.
A fix is under review in #2968 .

tuliom (approved):

mscastanho (approved):

PR #2930, opened 2022-06-21
Fix the content of the release package (SLES)

er-1:
Updated the name of the product based on this comment[1].
The end of life in the spec file wasn't correct. Fixed this
and added an AT variable with the end of life for each AT version.

[1] https://bugzilla.linux.ibm.com/show_bug.cgi?id=193326#c11

### DIFF
--- a/configs/13.0/specs/release.spec
+++ b/configs/13.0/specs/release.spec
@@ -13,9 +13,12 @@
 # limitations under the License.
 #
 
-# Per convention the %%product name will be used a stem for the -release
-# package and the -prod file
-%define         product advance-toolchain-%{at_major}
+# The product is called 'ibm-power-advance-toolchain' in SCC.
+%define         product ibm-power-advance-toolchain
+
+# Per convention the %%package name will be used a stem for the -release
+# package.
+%define         package advance-toolchain-%{at_major}
 
 # Decide: The products version and release often follows the release package.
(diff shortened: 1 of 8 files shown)


configs/13.0/specs/release.spec
```diff
@@ -27,15 +30,15 @@
...
 %define         producturlbugtracker https://github.com/advancetoolchain/advance-toolchain/issues/
 %define         producturlbugtracker_pcescaped https%3A%2F%2Fgithub.com%2Fadvancetoolchain%2Fadvance%2Dtoolchain%2Fissues%2F
 
 # ----------------------------------------------------------------------
 
-Name:           %{product}-release
+Name:           %{package}-release
 Summary:        Advance Toolchain
 Version:        %{productversion}
 Release:        %{productrelease}
```
tuliom, line 44: AFAIU, the way this is written could allow for a `-release` package from version 15.0 be installed without having an AT 15.0 installed.
Should we add a line around line 60 with the following?
```
Suggests: advance-toolchain-%{at_major}-runtime
```

This won't enforce the version, but would help the resolver to get the latest one.

tuliom (commented):
I have a suggestion. What do you think about this?

configs/13.0/specs/release.spec
er-1: Done

PR #2981, opened 2022-07-23
Update gcc on AT next

Cherbot:
Bump to revision b563a8dd3fcd


mscastanho:
The issue only happens when libgfortran is compiled with `-lto` (one of the patches we apply to `gcc_4`), that's why I wasn't able to reproduce it before. I just opened a bug report upstream: [PR106438](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=106438)

er-1:
Closing. #3186 will try with a newer revision.

PR #3001, opened 2022-08-02
Remove private URL from documentation

### LINKED ISSUE #2108: README: Outdated URL
The URL below is non-functional.

https://github.com/advancetoolchain/advance-toolchain/blob/d7a46e1af663f1283ac25dc982523fc48cd90266/README.txt#L176

    $ curl -I https://ltc3.linux.ibm.com/wiki/LoP_Toolchain/Advance_Toolchain/Education/ConfigSets
    curl: (6) Could not resolve host: ltc3.linux.ibm.com

er-1:
In the `README.txt` there is an URL pointing to an IBM's private network.
This patch removes it.

Fix #2108

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -174,8 +174,6 @@ Each and every configset must have this structure/files to assure that the build
 - sanity.mk file
   This file contains a macro definition to perform general sanity checks at the beginning of the build process. It should check for general requirements on the build system in order to run the build. Any package or program check required by the build should be set on the distro specific sanity check, not here.
 
-This covers our overview of configsets. If your are interested in more information about the configset structure, purpose and content, take a look here: https://ltc3.linux.ibm.com/wiki/LoP_Toolchain/Advance_Toolchain/Education/ConfigSets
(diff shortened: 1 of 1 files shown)


rzinsly (approved):

ThinkOpenly (approved):

paulmenzel:
Why can’t the information on that private/internal page be made public?

tuliom:
> Why can’t the information on that private/internal page be made public?

@paulmenzel That URL does not exist.
Anyway, I managed to find a copy of the wiki page this URL was trying to link and I published it here:
https://gist.github.com/tuliom/ee1eca9d04ec5a08deda911b6fda3279

As you can see, the contents are out-of-date.
This file also overlaps with our README.
Our direction now is to improve the README and source code comments related to the variables.  They're easier to maintain and to use when an explanation is needed.

paulmenzel:
Thank you for the investigation and explanation. Makes sense.

PR #3021, opened 2022-08-11
Disable cachemanager service setup on containers

fabiane-watanabe:
Podman/docker containers don't necessarily have systemd enabled so
AT installation on such systems fails when trying to setup the cachemanager
service. Simply don't try to setup the service when inside a container for now.

### DIFF
--- a/configs/13.0/deb/monolithic/runtime.postinst
+++ b/configs/13.0/deb/monolithic/runtime.postinst
@@ -24,7 +24,8 @@ if [[ "${1}" == configure ]]; then
 	[[ -f __AT_DEST__/sbin/ldconfig ]] && __AT_DEST__/sbin/ldconfig
 fi
 
-if [[ "__USE_SYSTEMD__" == "yes" && -n "$(command -v systemctl)" ]]; then
+if [[ "__USE_SYSTEMD__" == "yes" && -n "$(command -v systemctl)"\
+	&& "$(systemctl is-system-running)" != "offline" ]]; then
 	systemctl --no-reload preset __AT_VER_ALTERNATIVE__-cachemanager.service \
 	    > /dev/null 2>&1 || :
 	systemctl restart __AT_VER_ALTERNATIVE__-cachemanager.service || \
--- a/configs/13.0/deb/monolithic/runtime.prerm
+++ b/configs/13.0/deb/monolithic/runtime.prerm
@@ -17,7 +17,8 @@
 
 if [[ "${1}" == remove ]]; then
 	# Stop and disable the cache manager service to remove symlinks.
-	if [[ "__USE_SYSTEMD__" == "yes" && -n "$(command -v systemctl)" ]]; then
+	if [[ "__USE_SYSTEMD__" == "yes" && -n "$(command -v systemctl)"\
+		&& "$(systemctl is-system-running)" != "offline" ]]; then
 		systemctl stop __AT_VER_ALTERNATIVE__-cachemanager.service
 		systemctl disable __AT_VER_ALTERNATIVE__-cachemanager.service
 	fi
--- a/configs/14.0/specs/monolithic.spec
+++ b/configs/14.0/specs/monolithic.spec
@@ -263,7 +263,7 @@ fi
 # Automatically set the timezone
 rm -f %{_prefix}/etc/localtime
 ln -s /etc/localtime %{_prefix}/etc/localtime
-if [[ -n "$(command -v systemctl)" ]]; then
+if [[ -n "$(command -v systemctl)" && "$(systemctl is-system-running)" != "offline" ]]; then
 	systemctl preset %{at_ver_alternative}-cachemanager.service \
 	    > /dev/null 2>&1 || :
 	systemctl restart %{at_ver_alternative}-cachemanager.service
@@ -433,9 +433,10 @@ fi
 
 #---------------------------------------------------
 %preun runtime
-command -v systemctl >/dev/null && systemctl --no-reload disable --now \
+if [[ -n "$(command -v systemctl)" && "$(systemctl is-system-running)" != "offline" ]]; then
+    systemctl --no-reload disable --now \
     %{at_ver_alternative}-cachemanager.service > /dev/null 2>&1 || :
-
+fi
 ####################################################
 %postun runtime
 # Remove the directory only when uninstalling
@@ -452,9 +453,11 @@ if file /usr/sbin/ldconfig | grep "bash script" > /dev/null; then
         rm -f /usr/sbin/ldconfig
     fi
 fi
-command -v systemctl >/dev/null &&  systemctl try-restart %{at_ver_alternative}-cachemanager.service >/dev/null \
-    2>&1 || :
 
+if [[ -n "$(command -v systemctl)" && "$(systemctl is-system-running)" != "offline" ]]; then
+    systemctl try-restart %{at_ver_alternative}-cachemanager.service >/dev/null \
+    2>&1 || :
+fi
 #---------------------------------------------------
 %postun devel
 if [ "$1" = 0 ]; then


mscastanho (changes requested):
The changes look OK. 

This will effectively disable the `cachemanager` service setup when AT packages are installed on containers. Ideally we'd need another way to update AT's ld.so.cache on containers to guarantee everything works properly if the user installs/removes more libraries. Unfortunately, I can't think of a simple alternative at the moment. Still, I think this PR is at least a step further in the direction of supporting AT usage on containers.

But I believe this deserves a better commit message / description. The way it is it's just describing the issue, not the solution or why the choice for it. A better description will greatly help other future AT developers understand why this change was done. What about something like:

```
Disable cachemanager service setup on containers

podman/docker containers don't necessarily have systemd enabled so
AT installation on such systems fails when trying to setup the cachemanager
service. Simply don't try to setup the service when inside a container for now.
```

fabiane-watanabe:
@mscastanho I updated the commit message. Tks

configs/13.0/deb/monolithic/runtime.postinst
```diff
@@ -24,7 +24,8 @@ if [[ "${1}" == configure ]]; then
 	[[ -f __AT_DEST__/sbin/ldconfig ]] && __AT_DEST__/sbin/ldconfig
 fi
 
-if [[ "__USE_SYSTEMD__" == "yes" && -n "$(command -v systemctl)" ]]; then
+if [[ "__USE_SYSTEMD__" == "yes" && -n "$(command -v systemctl)"\
+&& ! -e '/.dockerenv' && ! -e '/run/.containerenv' ]]; then
```
ThinkOpenly, line 28: This line should be indented, probably such that the `&&` lines up with `"__USE_SYSTEMD__".

tuliom (approved):

ThinkOpenly (changes requested):
Some light Googling seems to imply that relying on the existence of `/.dockerenv` is fragile. I saw one recommendation of `/run/.dockerenv` instead, but I'm not sure how solid it is. Can we test for the _issue_ that is presented by being in a container, rather than just "being in a container"?

fabiane-watanabe:
> Some light Googling seems to imply that relying on the existence of `/.dockerenv` is fragile. I saw one recommendation of `/run/.dockerenv` instead, but I'm not sure how solid it is. Can we test for the _issue_ that is presented by being in a container, rather than just "being in a container"?


The issue happens in some distro where systemctl command is detected,
but its execution fails. 
In ubi9, systemctl is detected and occurs the following error when it´s executed: 
```
System has not been booted with systemd as init system (PID 1). Can't operate.
Failed to connect to bus: Host is down
```
And I am not sure if this error is restricted to container environment to be used as verification.
Since `./dockerenv` (docker) and `/run/.containerenv` (podman) are created automatically into container, I think it could be safe for now.
( I couldn´t find `/run/.dockerenv` in docker container, only `/.dockerenv` )

ThinkOpenly:
> > Some light Googling seems to imply that relying on the existence of `/.dockerenv` is fragile. I saw one recommendation of `/run/.dockerenv` instead, but I'm not sure how solid it is. Can we test for the _issue_ that is presented by being in a container, rather than just "being in a container"?
> 
> The issue happens in some distro where systemctl command is detected, but its execution fails.

Is it fair to say that the issue is that invoking `systemctl` fails?

If so, could we just run `systemctl` and check if it's working?
```
system$ systemctl >&/dev/null; echo $?
0
```
```
container$ systemctl >&/dev/null; echo $?
1
```

fabiane-watanabe:
> Is it fair to say that the issue is that invoking `systemctl` fails?
> 
> If so, could we just run `systemctl` and check if it's working?
> 
> ```
> system$ systemctl >&/dev/null; echo $?
> 0
> ```
> 
>  ```
> container$ systemctl >&/dev/null; echo $?
> 1
> ```

If systemctl fails but it's not an container? 
Since in a build of AT docker,  ldconfig command used is the AT ldconfig itself, I think there's no harm in these failure. 
`https://github.com/advancetoolchain/at-docker/blob/de150e5f8ccdf1f780c2bfc806853ab30eea5d40/dockerfile.template#L47`
But, if it's not a container, it's better warning the user indicating that cache manager setup fails.

ThinkOpenly:
> > Is it fair to say that the issue is that invoking `systemctl` fails?
> > If so, could we just run `systemctl` and check if it's working?
> 
> If systemctl fails but it's not an container?
> Since in a build of AT docker, ldconfig command used is the AT ldconfig itself, I think there's no harm in these failure. 
> But, if it's not a container, it's better warning the user indicating that cache manager setup fails.

There is another subcommand:
```
system$ systemctl is-system-running
running
```
```
container$ systemctl is-system-running
offline
```

Perhaps that should be used as the primary test, and then use the sketchier tests to determine whether to emit a warning? The above tests also avoid the generic "`systemctl` just fails" false positive result for "am I in a container?".

configs/13.0/deb/monolithic/runtime.postinst
fabiane-watanabe: fixed

configs/13.0/deb/monolithic/runtime.prerm
```diff
@@ -17,7 +17,8 @@
 
 if [[ "${1}" == remove ]]; then
 	# Stop and disable the cache manager service to remove symlinks.
-	if [[ "__USE_SYSTEMD__" == "yes" && -n "$(command -v systemctl)" ]]; then
+	if [[ "__USE_SYSTEMD__" == "yes" && -n "$(command -v systemctl)"\
+	&& ! -e '/.dockerenv' && ! -e '/run/.containerenv'  ]]; then
```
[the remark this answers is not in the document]
fabiane-watanabe: fixed

fabiane-watanabe:
> > > Is it fair to say that the issue is that invoking `systemctl` fails?
> > > If so, could we just run `systemctl` and check if it's working?
> > 
> > 
> > If systemctl fails but it's not an container?
> > Since in a build of AT docker, ldconfig command used is the AT ldconfig itself, I think there's no harm in these failure.
> > But, if it's not a container, it's better warning the user indicating that cache manager setup fails.
> 
> There is another subcommand:
> 
> ```
> system$ systemctl is-system-running
> running
> ```
> 
>  ```
> container$ systemctl is-system-running
> offline
> ```
> 
> Perhaps that should be used as the primary test, and then use the sketchier tests to determine whether to emit a warning? The above tests also avoid the generic "`systemctl` just fails" false positive result for "am I in a container?".


yes, I agree that this test is more container agnostic.
So, I will replace the previous test by this one:
 if [ "$(systemctl is-system-running)" != "offline" ];  then
        show error
 fi

tuliom:
> So, I will replace the previous test by this one:
> if [ "$(systemctl is-system-running)" != "offline" ]; then
> show error
> fi

@fabiane-watanabe What do you mean with "show error"?
Do you actually mean "run other systemctl" commands?

fabiane-watanabe:
> > So, I will replace the previous test by this one:
> > if [ "$(systemctl is-system-running)" != "offline" ]; then
> > show error
> > fi
> 
> @fabiane-watanabe What do you mean with "show error"? Do you actually mean "run other systemctl" commands?

Yes, it will continues the execution as it was.

configs/14.0/specs/monolithic.spec
```diff
@@ -433,9 +433,10 @@ fi
 
 #---------------------------------------------------
 %preun runtime
-command -v systemctl >/dev/null && systemctl --no-reload disable --now \
+if [[ -n "$(command -v systemctl)" && "$(systemctl is-system-running)" != "offline" ]]; then
+    systemctl --no-reload disable --now \
     %{at_ver_alternative}-cachemanager.service > /dev/null 2>&1 || :
-
+fi
```
tuliom, line 439: There are some trailing whitespaces here.

tuliom (approved):
The patch LGTM.
I found a minor cosmetic issue that I think that can be merged without a new test.

ThinkOpenly (approved):
[_agree with Tulio's comment_]

mscastanho (approved):

PR #3036, opened 2022-08-17
Update libnxz on AT 15.0 and 16.0

tuliom:
Let AT 16.0 reuse the source file from AT 15.0.


rzinsly (approved):

er-1:
`libnxz` can't be built for AT 15.0. I wonder if the `stage_1` file for `libnxz` in AT 15.0 doesn't match with latest version of `libnxz`.

er-1:
> `libnxz` can't be built for AT 15.0. I wonder if the `stage_1` file for `libnxz` in AT 15.0 doesn't match with latest version of `libnxz`.

I gave a try. `libnxz` in AT 15.0 with the `stage_1` file from AT 16.0, it passed.

mscastanho:
@er-1 Indeed, AT 15.0 was using libnxz v0.62, which didn't use Automake yet. So the steps to build it changed on v0.63. @tuliom If we are going to upgrade the libnxz version on AT 15, we need to update the `stage_1` file as well.

tuliom:
Fixed. Let's see if the this new version works.

PR #3103, opened 2022-10-08
Update gdb on AT 16.0

Cherbot:
Bump to revision dbac5b57d82f


er-1:
It failed on Debian 11 while testing `ck_provides`.

PR #3150, opened 2022-11-14
Remove config for AT 12.0

er-1:
There are no plans to update AT 12.0.
Removed config and refences to AT 12.0.


ThinkOpenly (changes requested):
Does this need to stay as-is?
```
./docs/advtool-usage:   <li>On Advance Toolchain 10.0, 11.0, 12.0, and 13.0 the compiler defaults to <code>-mcpu=power8 -mtune=power9</code></li>
```

Does this need to stay?
```
./configs/13.0/base.mk:AT_PREVIOUS_VERSION := 12.0
```

er-1:
> Does this need to stay as-is?
> 
> ```
> ./docs/advtool-usage:   <li>On Advance Toolchain 10.0, 11.0, 12.0, and 13.0 the compiler defaults to <code>-mcpu=power8 -mtune=power9</code></li>
> ```

I added the missing AT notes, for AT > 13.0.
I think the notes for unsupported versions can stay in the documentation as the packages are still available in the DHE repository.

> Does this need to stay?
> 
> ```
> ./configs/13.0/base.mk:AT_PREVIOUS_VERSION := 12.0
> ```

It's used by packaging scripts, even if the previous config is not available.
We might have to update later the scripts to run without the `AT_PREVIOUS_VERSION` variable.

ThinkOpenly (approved):

PR #3346, opened 2023-04-15
Update gdb on AT next

Cherbot:
Bump to revision 672008e6dcac


er-1:
Support of Debian 12 is ready yet.

Issue #3550, opened 2023-08-03
Update glibc to 2.38 on AT 17.0


er-1:

er-1:
#3551 fixes this.

PR #3551, opened 2023-08-03
Diverge glibc on AT 17.0 from AT next

er-1:
glibc 2.38 has been released and will be distributed on AT 17.0. AT next will continue to track glibc master branch.


er-1:
This fixes #3550

PR #3618, opened 2023-09-02
Update gdb on AT next

Cherbot:
Bump to revision 41770089162e


er-1:

PR #3619, opened 2023-09-02
Update glibc on AT next

Cherbot:
Bump to revision d2123d68275a


er-1:
Let's see if it'll pass with the weekly update.

PR #3621, opened 2023-09-02
Update binutils on AT 17.0

Cherbot:
Bump to revision ba510ea7710c


er-1:

PR #3647, opened 2023-09-14
Remove config for AT 13.0

er-1:
There are no plans to update AT 13.0.
Removed config and references to AT 13.0.


er-1:
The builds on AT 13.0 failed. That was expected.

PR #3741, opened 2023-10-28
Update gcc on AT next

Cherbot:
Bump to revision 7f974c5fd4f5


er-1:
Let's see how it goes with the next weekly update.

PR #3768, opened 2023-11-11
Update glibc on AT next

Cherbot:
Bump to revision d1dcb565a1fb


er-1:
Let's see if it passes with the next weekly update.

PR #3783, opened 2023-11-30
Update rpm building script

er-1:
On RHEL9 the script to collect debug information is not in the same directory as on RHEL8. This script is called when making the debug packages for AT.

With this patch, the correct file will used depending on the current RHEL version.


er-1:
The issue on SLES15 isn't related with this PR.

PR #3811, opened 2023-12-12
Update python on AT

er-1:
On AT < 16.0 and AT next, update the patch about `zlib` version check.
On AT 16.0 and AT 17.0, remove the patch about `zlib` version check.


er-1:
Close #3802
Close #3807

PR #4897, opened 2026-06-25
Update AT next

er-1:


er-1:
Debian 13, RHEL 10, SLES 16 and Ubuntu 26.04 are not yet supported, so it's expected to fail against those distros.
In addition, since the buildbot master has been moved to a new machine, we have got errors when uploading logs to the at-logs repository (it needs to be reset).
This means the PR can be merged. The errors are side effects.

PR #4901, opened 2026-06-29
Update gdb on AT next

Cherbot:
Bump to revision 0e73a13ca62a


er-1:

PR #4902, opened 2026-06-29
Update glibc on AT next

Cherbot:
Bump to revision ccd4cd5324ca


er-1:

